PR #41, opened 2024-09-05
Serve models

favyen2:
Start exploring how to serve models.


favyen2:
Closing since we decided to implement this differently.

PR #78, opened 2024-10-10
Add Molmo backbone

favyen2:


favyen2:
@yawenzzzz I fixed the issue with CLIP model for other variants and added test, can you take another look? thanks!

rslearn/models/clip.py
```diff
@@ -0,0 +1,62 @@
...
+
+
+class CLIP(torch.nn.Module):
+    """CLIP image encoder."""
+
+    def __init__(
+        self,
+        model_name: str,
+    ):
+        """Instantiate a new Molmo instance.
```
yawenzzzz, line 16: Change "Molmo" to "Clip"

yawenzzzz (approved):
LGTM, thanks for the fix!

PR #80, opened 2024-10-10
Improve metrics in DetectionTask.

favyen2:
Specifically, report precision/recall in addition to F1, and support reporting the F1 at multiple score thresholds.

We really need to figure out how to log dict returned by one metric though, right now it is convoluted (creating many classes) and inefficient (each class is basically repeating the same computation but just returning a different metric).

I think that can go in later PR but this one should have a test. So WIP for now.


favyen2:
@yawenzzzz I added you since this is needed for the change at https://github.com/allenai/rslearn_projects/pull/72

yawenzzzz (approved):
LGTM.

PR #85, opened 2024-10-18
Fix error when loading all patches with windows that have bands at lower resolution

favyen2:
In rslearn.train.Dataset, load_all_patches means we load all patches of a fixed size that together cover the window. This means sometimes the patches can be close to the border of the window, e.g. if window bounds is [0, 0, 257, 257] with 256x256 patch size then one patch would be [256, 256, 512, 512]. This is fine but if there's a lower resolution band (negative zoom offset) it could end up being out of bounds.

Previously the GeotiffRasterFormat retuned None if it is out of bounds for some reason, so this just fixes it to return a zero-valued array of the appropriate # bands and dtype.


yawenzzzz (approved):
LGTM.

PR #88, opened 2024-10-21
Some changes to speed up the prediction pipeline in reslearn_projects

yawenzzzz:
1. Apply `restore_config` only for training, in this way, using the same configuration file, we won't apply `restore_config` when doing validation, test, and predict.
2. Add cache decorator to `data_source_from_config` to store the call results in memory.


favyen2 (approved):
cool, this is much better!

PR #91, opened 2024-10-22
Support loading time series layer without adding extra layers to dataset config

### LINKED ISSUE #86: Improve time series handling
Currently if we have a layer X that has `query_config.max_matches > 1` then they are stored into these layers named X, X.1, X.2, etc.

In the model configuration file you need to refer to them as X, X.1, X.2, etc.

This is all okay but during training, when the dataset sees "X.1" it looks for that layer in the dataset config file, meaning we need "X.1" and "X.2" and so on to then be defined in the dataset config.

So at the very least we should make it so that it auto-detects that X.1 should just be X. Which could be as simple as reserving "." in the layer name and splitting it. Or could see if there are other problems that can be solved at the same time (I don't know if anything more is necessary actually).

favyen2:
Support loading time series layer without adding extra layers to dataset config

Resolves #86


favyen2:
I will try to add test for this after #81 is merged, so marking draft for now.

rslearn/dataset/dataset.py
```diff
@@ -48,10 +48,13 @@ def __init__(self, path: UPath) -> None:
...
             config = json.load(f)
-            self.layers: dict = {
-                layer_name: load_layer_config(d)
-                for layer_name, d in config["layers"].items()
-            }
+            self.layers = {}
+            for layer_name, d in config["layers"].items():
+                # Layer names must not contain period, since we use period to
+                # distinguish different materialized groups within a layer.
+                assert "." not in layer_name
```
yawenzzzz, line 55: It will be better to add a message to this assert statement, like what's described in the comment

yawenzzzz (approved):
LGTM, just one minor comment.

rslearn/dataset/dataset.py
favyen2: thanks, added!

PR #93, opened 2024-10-23
Henryh/add lint workflows

Hgherzog:
- I chose to Raise an error if something was unexpectedly None in most cases unless another option seemed clear
- I added a bunch of TODOs and notes as I went through that I decided were out of scope of this code heavy change PR


rslearn/models/swin.py
```diff
@@ -141,6 +144,7 @@ def forward(
                 process.
             targets: target dicts that are ignored unless
         """
+        # TODO: Are Targets  supposed to be ignored? Is this to maintain a common interface?
```
favyen2, line 147: Hm I think this was so that the module could be used both as an "encoder" in MultiTask / SingleTask or as the full classification model, but really for classification it would make sense to pair it (with output_layers=None) with ClassificationHead so I think just taking inputs makes sense since in that case it would still be used as "encoder" (which is just passed the inputs). So I'll change this.

rslearn/train/prediction_writer.py
```diff
@@ -154,7 +160,8 @@ def write_on_batch_end(
 
             # This is the last patch so it's time to merge outputs from overlapped patches
             if self.merger is not None:
-                pending_output = self.merger.merge(pending_output)
+                # TODO: HackyNot sure why we don't pass metadata to the merger can't see any subclasses
```
favyen2, line 163: I updated the PatchPredictionMerger to just take output since that matches rslp.utils.nms.NMSDistanceMerger.

rslearn/train/tasks/classification.py
```diff
@@ -120,9 +120,12 @@ def process_inputs(
 
         data = raw_inputs["targets"]
         for feat in data:
-            for property_name, property_value in self.filters:
-                if feat.properties.get(property_name) != property_value:
-                    continue
+            if feat.properties is None:
+                continue
+            if self.filters is not None:
```
favyen2, line 125: I can't remember why filters was allowed to be None but I don't see a good reason for it now, so I changed the __init__ argument type / default.

rslearn/utils/mp.py
```diff
@@ -33,7 +34,7 @@ def star_imap_unordered(
     p: multiprocessing.pool.Pool,
     fn: Callable[..., Any],
     kwargs_list: list[dict[str, Any]],
-) -> Generator[Any, None, None]:
+) -> IMapIterator[Any]:
```
favyen2, line 37: Should this just be `IMapIterator` (without subscript)? I get error:
```TypeError: type 'IMapIterator' is not subscriptable```

File: tests/integration/data_sources/test_gcp_public_data.py [hunk 5]
```diff
@@ -18,12 +19,15 @@
 from rslearn.utils import STGeometry
 
 
+# What functionality are we really testing here how does local vs gcs compare is it the ingestion Otherwise we can do in memory bucket but unsure
```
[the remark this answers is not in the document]
favyen2: We can change this later.

favyen2:
I updated some things that I thought made sense to fix now rather than later. Can you look into these?
* IMapIterator[Any]
* I think comment about "# TODO: NOT SURE This is correct behavior or we should return None" can be removed in `data_sources/copernicus.py`.
* Unused  `copy_cache_to_local_dir` function in `utils/rtree_index.py`.
* Defining `parent_file` in `rslearn/utils/raster_format.py` to create the parent directory, but it should've already been created on the line before.

rslearn/utils/raster_format.py
```diff
@@ -339,7 +339,11 @@ def encode_raster(
             profile["blockysize"] = self.block_size
 
         path.mkdir(parents=True, exist_ok=True)
-        with (path / self.fname).open("wb") as f:
+        path_to_write = path / self.fname
+        parent_file = path_to_write.parent
+        parent_file.mkdir(parents=True, exist_ok=True)
```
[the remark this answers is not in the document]
Hgherzog: I just did this when I was debugging that fake gcs server I will change it back

favyen2 (approved):

PR #96, opened 2024-10-25
Improved GeoJSON support

### LINKED ISSUE #26: Support flag for GeoJSON vector format to use longitude/latitude coordinates

favyen2:
Improve GeoJSON support:
* GeojsonVectorFormat - support a "coordinate mode" option to write GeoJSON with CRS coordinates. Previously it writes "pixel coordinates" which are CRS coordinates multiplied by x/y resolution. That option is still there, but the default will now be to write CRS coordinates since these are understood by other GIS tools. Information about which coordinates are used is included in the GeoJSON file, so GeojsonVectorFormat can always read previously written files even if they don't match the configured coordinate mode.
* GeojsonVectorFormat - support option to always write longitude/latitude (WGS84) coordinates.
* LocalFiles - instead of always assuming input GeoJSON is longitude/latitude, interpret the "crs" property if it is set. This is all handled by fiona. Note that GeoJSON created by GeojsonVectorFormat in pixel coordinates won't be interpreted correctly by LocalFiles.

Resolves #26.


favyen2:
todo: unit test for GeojsonVectorFormat

rslearn/data_sources/local_files.py
```diff
@@ -353,14 +360,16 @@ def ingest_item(
                 aux_files.append(path.parent / (prefix + ext))
 
         # TODO: move converting fiona file to list[Feature] to utility function.
-        # TODO: don't assume WGS-84 projection here.
         with get_upath_local(path, extra_paths=aux_files) as local_fname:
             with fiona.open(local_fname) as src:
+                crs = CRS.from_wkt(src.crs.to_wkt())
+                projection = Projection(crs, 1, 1)
```
Hgherzog, line 366: Suggestion: Do we want the resolution to always be hardcoded as 1? If yes, prefer using the keywords here for readability. If not should be configurable.

rslearn/utils/vector_format.py
```diff
@@ -180,12 +182,81 @@ def from_config(name: str, config: dict[str, Any]) -> "TileVectorFormat":
...
+                    "name": output_projection.crs.to_wkt(),
+                },
+            }
+
+        fc["features"] = []
+        for feat in features:
+            feat = feat.to_projection(output_projection)
+            fc["features"].append(feat.to_geojson())
+
+        with fname.open("w") as f:
```
Hgherzog, line 257: I think it is a good convention to log what we are writing and where when we write files. Makes it easier for someone using rslearn to understand what is happening

tests/integration/data_sources/test_local_files.py
```diff
@@ -32,3 +44,108 @@ def test_sample_dataset(self, local_files_dataset: Dataset) -> None:
...
+        good_window.save()
+
+        dataset = Dataset(ds_path)
+
+        windows = dataset.load_windows()
+        prepare_dataset_windows(dataset, windows)
+        ingest_dataset_windows(dataset, windows)
+        materialize_dataset_windows(dataset, windows)
+
+        assert not (bad_window.path / "layers" / "local_file" / "data.geojson").exists()
```
Hgherzog, line 150: Suggestion: If it is not time restricting, it would be better to have a seperate test for each assert so if one fails you know more clearly in which case the functionality is broken.

tests/integration/data_sources/test_local_files.py
```diff
@@ -32,3 +44,108 @@ def test_sample_dataset(self, local_files_dataset: Dataset) -> None:
...
+            Feature(
+                geometry=seattle.to_projection(custom_projection),
+            ),
+        ]
+        src_data_dir = ds_path / "src_data"
+        src_data_dir.mkdir(parents=True, exist_ok=True)
+        vector_format = GeojsonVectorFormat(coordinate_mode=coordinate_mode)
+        vector_format.encode_vector(src_data_dir, custom_projection, features)
+
+        # Primary purpose of this test is to make sure LocalFiles can import the data,
```
Hgherzog, line 70: Suggestion: Move this part where you verify the encodding based on the desired coordinate mode to a separate integration test. It is clearer and easier to debug tests that only  have a single assert (or parametrized set of asserts. Also, consider having a different test for each coordinate mode as that will make things more readable.

tests/integration/data_sources/test_local_files.py
```diff
@@ -32,3 +44,108 @@ def test_sample_dataset(self, local_files_dataset: Dataset) -> None:
...
+            projection=window_projection,
+            bounds=(
+                int(window_center.x) - 10,
+                int(window_center.y) - 10,
+                int(window_center.x) + 10,
+                int(window_center.y) + 10,
+            ),
+            time_range=None,
+        )
+        good_window.save()
```
Hgherzog, line 141: Suggestion: Move the good and bad window to pytest fixtures if possible as well as the dataset config

tests/unit/utils/test_vector_format.py
```diff
@@ -0,0 +1,83 @@
...
+from rasterio.crs import CRS
+from upath import UPath
+
+from rslearn.const import WGS84_PROJECTION
+from rslearn.utils.feature import Feature
+from rslearn.utils.geometry import Projection, STGeometry
+from rslearn.utils.vector_format import GeojsonCoordinateMode, GeojsonVectorFormat
+
+
+@pytest.mark.parametrize(
```
Hgherzog, line 16: Suggestion: Parametrize is ideal imo when you can specify both the inputs and outputs within the same logic. If you need separate logic for each case I would recommend just having separate unit tests so it is more readable.

Hgherzog (approved):
Approved with some minor comments about structuring the tests and ideas on places to add logging

tests/integration/data_sources/test_local_files.py
favyen2: I'll remove this section, I guess it is already covered by the unit test.

PR #97, opened 2024-10-29
fix workflow typo

Hgherzog:


Hgherzog:
https://github.com/allenai/rslearn/actions/runs/11578427684 Manual trigger of the workflow showing the tests do actually launch

PR #99, opened 2024-11-04
Henryh/favyen/planetbasemap

Hgherzog:
-Add planet Basemap and integration test
-allow disabled layers flag for prepare/ingest/materialize workflow
-add some logging when we write files places


Hgherzog:
https://github.com/allenai/rslearn/actions/runs/11674863750 integration test run

rslearn/log_utils.py
```diff
@@ -0,0 +1,22 @@
+"""Logging utilities."""
+
+import logging
+import sys
+
+LOG_FORMAT = "format=%(asctime)s loglevel=%(levelname)-6s logger=%(name)s %(funcName)s() L%(lineno)-4d %(message)s"
+# DETAILED_LOG_FORMAT = "format=%(asctime)s loglevel=%(levelname)-6s logger=%(name)s %(funcName)s() L%(lineno)-4d %(message)s   call_trace=%(pathname)s L%(lineno)-4d"  # noqa
+
+
+def get_logger(name: str) -> logging.Logger:
```
favyen2, line 10: This seems to break the RSLEARN_LOGLEVEL environment variable. Instead of always setting to DEBUG, can you set it based on the environment variable like it does currently?

rslearn/main.py
```diff
@@ -525,6 +542,12 @@ def dataset_ingest() -> None:
         prog="rslearn dataset ingest",
         description="rslearn dataset ingest: ingest items in retrieved data sources",
     )
+    parser.add_argument(
+        "--disabled-layers",
+        type=lambda s: s.split(","),
+        default="[]",
```
favyen2, line 548: Should the default here be empty string? I think the processing function should then return empty array if the input is empty string. Otherwise it works for this disabled layer since the layer "[]" probably doesn't exist but it would be confusing pattern. At least that is my understanding of what is happening here.

favyen2: Or maybe the default should be [] without the string.

favyen2 (approved):

PR #105, opened 2024-11-18
Support using ZSTD compression in GeoTIFFs

favyen2:
This PR depends on #104 since it changes functionality in the new DefaultTileStore.

Currently we use LZW compression only.

This PR adds geotiff_options to GeotiffRasterFormat and DefaultTileStore so that arbitrary rasterio profile options can be overwritten. Then we can use the geotiff_options to enable ZSTD compression. In particular this seems to give a good tradeoff between speed and compression ratio:

```
raster_format = GeotiffRasterFormat(geotiff_options=dict(compress="zstd", zstd_level=1, predictor=2))
```


Hgherzog (approved):
LGTM

PR #106, opened 2024-11-19
fix bug with model test / model predict when using multiple GPUs

favyen2:
Because previously the ordering can differ across workers, when pytorch lightning tries to split the work across GPUs it splits the indexes of the dataset but the windows at those indexes can differ across the GPUs so some windows get processed multiple times while others aren't processed at all.

This adds sorting by hash of window name so the ordering is consistent.


yawenzzzz (approved):
LGTM. Thanks for the fix!

PR #107, opened 2024-11-20
v0.0.2 release

favyen2:
I think it makes sense to put a "v0.0.2" on pip before looking at PRs like #104 / #105 which have more substantial changes.

(e.g. it's possible that #104 will introduce subtle changes in how images are rendered and require retraining model)


yawenzzzz (approved):
LGTM.

Issue #103, opened 2024-11-13
Better programmatic access to operations like creating window, loading a particular layer from window, etc.


favyen2:
Currently we construct paths manually both when populating a dataset with windows, and when getting specific files we want out of the dataset (e.g. after running materialize or after running prediction).

It would be better if these were library calls.

PR #109, opened 2024-11-21
Add function to get directory of a materialized layer within a window

favyen2:
Addresses the most obvious TODO in #103 

Note this PR depends on #104.


rslearn/dataset/manage.py
```diff
@@ -198,8 +198,8 @@ def materialize_window(
         layer_cfg: the layer config
     """
     # Check if layer is materialized already.
-    completed_fname = window.path / "layers" / layer_name / "completed"
-    if completed_fname.exists():
+    layer_dir = window.get_layer_dir(layer_name)
+    if (layer_dir / "completed").exists():
```
Hgherzog, line 202: nit: perhaps have a is_complet utility function that just checks for a completed file

rslearn/dataset/window.py
```diff
@@ -13,6 +13,47 @@
 
 logger = get_logger(__name__)
 
+LAYERS_DIRECTORY_NAME = "layers"
```
Hgherzog, line 16: I would recommend adding some tests for these directory structure functions

Hgherzog (approved):
LGTM, a couple minor suggestions. Thanks for doing this it is really helpful!

rslearn/dataset/manage.py
favyen2: added

rslearn/dataset/window.py
favyen2: added a few tests. some of them will just throw error if we change the directory structure, while the one for mark_layer_completed/is_layer_completed will actually test logic.

PR #110, opened 2024-11-21
Address issue where a small number of Sentinel-2 scenes in GCP bucket are missing some bands

favyen2:
Previously we ignored the error during ingestion. Then we changed it so the error remains. Neither is perfect since this means either the job fails or some items failed to ingest; it would be better to avoid those scenes during prepare (get_items) so that we can read other windows instead (if possible).

This PR implements that, so in get_items we only match with scenes that have all the bands intact in the bucket.

Note this PR depends on #104.


rslearn/data_sources/gcp_public_data.py
```diff
@@ -295,25 +324,16 @@ def _get_xml_by_name(self, name: str) -> ET.ElementTree:
...
-        Args:
-            name: the name of the item to get
-
-        Returns:
-            the item object
-        """
+    def _get_item_by_name(self, name: str) -> Sentinel2Item:
+        # Get the item without the caching logic.
         parts = name.split("_")
         assert len(parts[5]) == 6
```
Hgherzog, line 338: if this is for debugging remove, otherwise I suggest making a seperate function that you call to raise an error which makes it clear why something like len(parts[5])  must be 6

rslearn/data_sources/gcp_public_data.py
```diff
@@ -347,6 +367,19 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
             raise ValueError(f"IMAGE_FILE is empty for {name}")
         blob_prefix = base_url + elements[0].text.split("B01")[0]
 
+        # Some Sentinel-2 scenes in the bucket are missing a subset of image files. So
```
Hgherzog, line 370: Suggest factoring this into a clearly named helper function to improve readability

rslearn/data_sources/gcp_public_data.py
```diff
@@ -362,13 +395,112 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
...
+            raise
+
+        with open_atomic(cache_item_fname, "w") as f:
+            json.dump(item.serialize(), f)
+        return item
+
+    def _read_products_for_cell_year(
+        self, cell_id: str, year: int
+    ) -> list[Sentinel2Item]:
+        # Read items for the given cell and year. These items are cached by
```
Hgherzog, line 451: add doc string

rslearn/data_sources/gcp_public_data.py
```diff
@@ -362,13 +395,112 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
...
+        # _read_products together in one file.
+        cell_part1 = cell_id[0:2]
+        cell_part2 = cell_id[2:3]
+        cell_part3 = cell_id[3:5]
+
+        items = []
+
+        for product_prefix in ["S2A_MSIL1C", "S2B_MSIL1C"]:
```
Hgherzog, line 459: Create a constant for this list

rslearn/data_sources/gcp_public_data.py
```diff
@@ -362,13 +395,112 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
...
+            geometry.shp = shapely.make_valid(geometry.shp)
+
         return Sentinel2Item(
             name,
             geometry,
             blob_prefix,
             cloud_cover,
         )
 
+    def get_item_by_name(self, name: str) -> Sentinel2Item:
```
Hgherzog, line 410: Consider splitting this into a couple smaller functions for readability

rslearn/data_sources/gcp_public_data.py
```diff
@@ -72,6 +72,35 @@ def deserialize(d: dict[str, Any]) -> "Sentinel2Item":
         )
 
 
+class CorruptItemException(Exception):
```
Hgherzog, line 75: We should have tests for each of these exceptions

Hgherzog (commented):
Overall looks good, need to add some tests for the error catching functionality, otherwise minor comments

rslearn/data_sources/gcp_public_data.py
favyen2: added tests

rslearn/data_sources/gcp_public_data.py
favyen2: revamped

rslearn/data_sources/gcp_public_data.py
favyen2: added

rslearn/data_sources/gcp_public_data.py
favyen2: revamped

rslearn/data_sources/gcp_public_data.py
favyen2: revamped

favyen2:
thanks, made these updates!

favyen2:
@Hgherzog can you review this with the updates?

rslearn/data_sources/gcp_public_data.py
```diff
@@ -295,33 +343,44 @@ def _get_xml_by_name(self, name: str) -> ET.ElementTree:
         if not cache_xml_fname.exists():
             metadata_blob_path = base_url + "MTD_MSIL1C.xml"
```
Hgherzog, line 344: Suggestion: Make the "MTD_MSIL1C.xml" string into a constant variable for readability.

rslearn/data_sources/gcp_public_data.py
```diff
@@ -295,33 +343,44 @@ def _get_xml_by_name(self, name: str) -> ET.ElementTree:
         if not cache_xml_fname.exists():
```
Hgherzog, line 343: Not related to this PR but the string manipulation above is quite hard to parse without additional context, perhaps having a local method or variable name that indicates what 
cell_id[0:2] should return and make claer what the assumptions on the naming conventions are if this were to change

rslearn/data_sources/gcp_public_data.py
```diff
@@ -295,33 +343,44 @@ def _get_xml_by_name(self, name: str) -> ET.ElementTree:
...
         """
+        # Parse the product name, e.g.
+        # S2B_MSIL1C_20211114T083119_N0301_R021_T36RUU_20211114T091848.
+        # We just need the cell ID from the name (here it is 36RUU) since that is part
+        # of the path in the GCS bucket.
         parts = name.split("_")
-        assert len(parts[5]) == 6
-        assert parts[5][0] == "T"
-        cell_id = parts[5][1:]
+        cell_id_with_prefix = parts[5]
```
Hgherzog, line 368: Suggestion: factor out the string parsing and data verification checks into a seperate function or class for readability so that the reader can understand what information you are pulling out without caring about how you manipulate the string

rslearn/data_sources/gcp_public_data.py
```diff
@@ -347,28 +406,166 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
             raise ValueError(f"IMAGE_FILE is empty for {name}")
         blob_prefix = base_url + elements[0].text.split("B01")[0]
 
+        # Get the sensing start time.
```
Hgherzog, line 409: Nitpick: Looks like pulling out "PRODUCT_START_TIME" and "Cloud_Coverage_Assessment" can be refactored into a single function though the return type str parsing would still need to be handled individually

rslearn/data_sources/gcp_public_data.py
```diff
@@ -347,28 +406,166 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
...
+
+    def _read_products_for_cell_year(
+        self, cell_id: str, year: int
+    ) -> list[Sentinel2Item]:
+        """Read items for the given cell and year directly from the GCS bucket.
+
+        This helper function is used by self._read_products which then caches the
+        items together in one file.
+        """
+        cell_part1 = cell_id[0:2]
```
Hgherzog, line 518: This cell id parsing logic seems to be all over, would be helpful to combine into some helper functions.

rslearn/data_sources/gcp_public_data.py
```diff
@@ -347,28 +406,166 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
...
+        cell_part2 = cell_id[2:3]
+        cell_part3 = cell_id[3:5]
+
+        items = []
+
+        for product_prefix in self.VALID_PRODUCT_PREFIXES:
+            cell_folder = f"tiles/{cell_part1}/{cell_part2}/{cell_part3}"
```
Hgherzog, line 525: Similarly building the folder from the parts could be placed into a helper function used all over

rslearn/data_sources/gcp_public_data.py
```diff
@@ -347,28 +406,166 @@ def get_item_by_name(self, name: str) -> Sentinel2Item:
...
+            blob_prefix = f"{cell_folder}/{product_prefix}_{year}"
+            blobs = self.bucket.list_blobs(prefix=blob_prefix, delimiter="/")
+
+            # Need to consume the iterator to obtain folder names.
+            # See https://cloud.google.com/storage/docs/samples/storage-list-files-with-prefix#storage_list_files_with_prefix-python # noqa: E501
+            # Previously we checked for .SAFE_$folder$ blobs here, but those do
+            # not exist for some years like 2017.
+            for _ in blobs:
```
Hgherzog, line 533: Nitpick: perhaps place a large upper bound in the off case a prefix with an insanely humongous amount of files by accidentally gets passed to this. list_blobs could potentially be very expensive if ran without being checked

Hgherzog (approved):
Generally looks good, some minor comments about path manipulation and some refactoring suggestions

rslearn/data_sources/gcp_public_data.py
favyen2: I think it may be better to have it all consolidated here rather than have a separate function for each step of the XML parsing.

rslearn/data_sources/gcp_public_data.py
favyen2: fixed

rslearn/data_sources/gcp_public_data.py
favyen2: I think it may be confusing/error-prone to add a limit, and the only way this can happen is if the GCS bucket actually contains so many folders for some reason. Also the bucket is free, it is public (not requester pays).

rslearn/data_sources/gcp_public_data.py
favyen2: moved building the cell folder to separate function

rslearn/data_sources/gcp_public_data.py
favyen2: fixed

PR #111, opened 2024-12-04
Add ERA5 monthly data

yawenzzzz:
For ERA5 data, I used the Climate Data Store (CDS) api: https://cds.climate.copernicus.eu/how-to-api. The ERA5-Land monthly averaged data from 1950 to present, specifically monthly averaged analysis data, link: https://cds.climate.copernicus.eu/datasets/reanalysis-era5-land-monthly-means?tab=download, has been added.

Since it's monthly data, I created one item per month, and its time_range covers the start and end date for a month. For a given STGeometry, it will generate all the year & month items within its time_range.

### DIFF
--- a/rslearn/data_sources/climate_data_store.py
+++ b/rslearn/data_sources/climate_data_store.py
@@ -190,9 +190,9 @@ class ERA5LandMonthlyMeans(DataSource):
                 raise ValueError(
                     f"Latitude spacing is not uniform: {lat[i + 1] - lat[i]}"
                 )
-        west = lon.min()
-        north = lat.max()
-        pixel_size_x, pixel_size_y = self.PIXEL_SIZE, -self.PIXEL_SIZE
+        west = lon.min() - self.PIXEL_SIZE / 2
+        north = lat.max() + self.PIXEL_SIZE / 2
+        pixel_size_x, pixel_size_y = self.PIXEL_SIZE, self.PIXEL_SIZE
         transform = from_origin(west, north, pixel_size_x, pixel_size_y)
         crs = f"EPSG:{WGS84_EPSG}"
         with rasterio.open(
--- a/rslearn/data_sources/climate_data_store.py
+++ b/rslearn/data_sources/climate_data_store.py
@@ -36,7 +36,7 @@ class ERA5LandMonthlyMeans(DataSource):
     PIXEL_SIZE = 0.1  # degrees, native resolution is 9km
 
     # see: https://confluence.ecmwf.int/display/CKB/ERA5-Land%3A+data+documentation
-    # Table 2 & 3 for variable shortnames
+    # For variable full & short names
 
     def __init__(
         self,
@@ -140,7 +140,7 @@ class ERA5LandMonthlyMeans(DataSource):
             # Variable data is stored in a 3D array (1, height, width)
             if len(band_data.shape) == 3:
                 bands_data.append(band_data[0, :, :])
-        tif_data = np.array(bands_data)  # shape (num_bands, height, width)
+        tif_data = np.array(bands_data)  # (num_bands, height, width)
         # Get metadata for the GeoTIFF
         lat = nc.variables["latitude"][:]
         lon = nc.variables["longitude"][:]
@@ -197,16 +197,16 @@ class ERA5LandMonthlyMeans(DataSource):
                 "year": [f"{item.geometry.time_range[0].year}"],  # type: ignore
                 "month": [
                     f"{item.geometry.time_range[0].month:02d}"  # type: ignore
-                ],  # Zero-padded month
-                "time": ["00:00"],  # default to 00:00
+                ],
+                "time": ["00:00"],
                 "area": [
                     bounds[3],
                     bounds[0],
                     bounds[1],
                     bounds[2],
-                ],  # [max_lat, min_lon, min_lat, max_lon]
-                "data_format": "netcdf",  # default to netcdf
-                "download_format": "unarchived",  # default to unarchived
+                ],
+                "data_format": "netcdf",
+                "download_format": "unarchived",
             }
             with tempfile.TemporaryDirectory() as tmp_dir:
                 local_nc_fname = os.path.join(tmp_dir, f"{item.name}.nc")
--- a/rslearn/data_sources/climate_data_store.py
+++ b/rslearn/data_sources/climate_data_store.py
@@ -2,6 +2,7 @@
 
 import os
 import tempfile
+from datetime import datetime, timezone
 
 import cdsapi
 import netCDF4
@@ -25,7 +26,7 @@ logger = get_logger(__name__)
 class ERA5LandMonthlyMeans(DataSource):
     """A data source for ingesting ERA5 land monthly averaged data from the Copernicus Climate Data Store.
 
-    The API key should be set via environment variable (CDS_API_KEY).
+    The API key should be set via environment variable (CDSAPI_KEY).
     """
 
     api_url = "https://cds.climate.copernicus.eu/api"
@@ -33,6 +34,8 @@ class ERA5LandMonthlyMeans(DataSource):
     # see: https://cds.climate.copernicus.eu/cdsapp#!/dataset/reanalysis-era5-land-monthly-means
     DATASET = "reanalysis-era5-land-monthly-means"
     PRODUCT_TYPE = "monthly_averaged_reanalysis"
+    DATA_FORMAT = "netcdf"
+    DOWNLOAD_FORMAT = "unarchived"
     PIXEL_SIZE = 0.1  # degrees, native resolution is 9km
 
     # see: https://confluence.ecmwf.int/display/CKB/ERA5-Land%3A+data+documentation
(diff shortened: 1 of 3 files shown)


rslearn/data_sources/climate_data_store.py
```diff
@@ -0,0 +1,219 @@
+"""Data source for Copernicus Climate Data Store."""
+
+import os
+import tempfile
+
+import cdsapi
```
favyen2, line 6: add to `extra_requirements.txt`

rslearn/data_sources/climate_data_store.py
```diff
@@ -0,0 +1,219 @@
...
+        nc = netCDF4.Dataset(nc_path)
+        bands_data = []
+        for band_name in nc.variables:
+            band_data = nc.variables[band_name]
+            # Variable data is stored in a 3D array (1, height, width)
+            if len(band_data.shape) == 3:
+                bands_data.append(band_data[0, :, :])
+        array = np.array(bands_data)  # (num_bands, height, width)
+        # Get metadata for the GeoTIFF
+        lat = nc.variables["latitude"][:]
```
favyen2, line 145: Are the latitudes and longitudes always evenly spaced by exactly 0.1 degrees? Otherwise this code may not work correctly.

favyen2: If it is not evenly spaced then you may need to use something like https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.griddata.html#scipy.interpolate.griddata to do bilinear interpolation onto a grid with fixed pixel size first, then compute the transform from the bounds of that interpolated grid.

rslearn/data_sources/climate_data_store.py
```diff
@@ -0,0 +1,219 @@
...
+        self.product_type = self.PRODUCT_TYPE
+
+        if api_key is None:
+            api_key = os.environ["CDS_API_KEY"]
+        # Follow CDS API docs: https://pypi.org/project/cdsapi/
+        # Copy both url and key to $HOME/.cdsapirc
+        with open(os.path.expanduser("~/.cdsapirc"), "w") as f:
+            f.write(f"url: {self.api_url}\nkey: {api_key}")
+
+        self.client = cdsapi.Client()
```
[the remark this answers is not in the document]
yawenzzzz: Good point, I have updated the code to pass the url and key directly.

rslearn/data_sources/climate_data_store.py
```diff
@@ -0,0 +1,219 @@
...
+
+        Returns:
+            A new ERA5LandMonthlyMeans instance
+        """
+        if config.data_source is None:
+            raise ValueError("data_source is required")
+        d = config.data_source.config_dict
+        return ERA5LandMonthlyMeans(
+            config=config,
+            api_key=d["api_key"],
```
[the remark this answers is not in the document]
yawenzzzz: Right, I have updated it to `d.get("api_key")` in case there's no `api_key` in the config.

rslearn/data_sources/climate_data_store.py
```diff
@@ -0,0 +1,219 @@
...
+            List of groups of items that should be retrieved for each geometry.
+        """
+        wgs84_geometries = [
+            geometry.to_projection(WGS84_PROJECTION) for geometry in geometries
+        ]
+        groups = []
+        for geometry in wgs84_geometries:
+            cur_items = []
+            # Create an item for each year & month within the time range
+            start_date = geometry.time_range[0].replace(day=1)  # type: ignore
```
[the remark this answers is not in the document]
yawenzzzz: Nice catch, yes, it's better to make hour/minute/second as 0 too, as we're using 00:00 later. I updated this with a new datetime for both start and end date.

rslearn/data_sources/climate_data_store.py
yawenzzzz: added!

rslearn/data_sources/climate_data_store.py
yawenzzzz: The data on CDS has been regridded to a regular lat-lon grid of 0.1° x 0.1° (reference: https://cds.climate.copernicus.eu/datasets/reanalysis-era5-land-monthly-means?tab=overview), I also downloaded an example nc file, here're how its latitude and longitude look like: 
`<xarray.DataArray 'latitude' (latitude: 61)> Size: 488B
array([46. , 45.9, 45.8, 45.7, 45.6, 45.5, 45.4, 45.3, 45.2, 45.1, 45. , 44.9,
       44.8, 44.7, 44.6, 44.5, 44.4, 44.3, 44.2, 44.1, 44. , 43.9, 43.8, 43.7,
       43.6, 43.5, 43.4, 43.3, 43.2, 43.1, 43. , 42.9, 42.8, 42.7, 42.6, 42.5,
       42.4, 42.3, 42.2, 42.1, 42. , 41.9, 41.8, 41.7, 41.6, 41.5, 41.4, 41.3,
       41.2, 41.1, 41. , 40.9, 40.8, 40.7, 40.6, 40.5, 40.4, 40.3, 40.2, 40.1,
       40. ])`
`<xarray.DataArray 'longitude' (longitude: 21)> Size: 168B
array([-116. , -115.9, -115.8, -115.7, -115.6, -115.5, -115.4, -115.3, -115.2,
       -115.1, -115. , -114.9, -114.8, -114.7, -114.6, -114.5, -114.4, -114.3,
       -114.2, -114.1, -114. ])`
which strictly have the 0.1° interval. 
Screenshot 2024-12-05 at 6 10 37 PM

favyen2: Ah ok great! Not absolutely necessary but minor suggestion to add a check that raises ValueError in case the spacing is not as expected.

yawenzzzz: make sense! I have added the spacing check for both latitude and longitude grids.

yawenzzzz:
> I think I still run into error when running `rslearn dataset ingest` because of deserialize_item not being implemented.

Ah, sorry, I added this the deserialize_item function. I made another change to speed up the ingest process, basically, in `get_items` instead of get items for each geometry separately, we union ST geometries first, and generate union items that can be used as candidate items for all geometries, and each item correspond to one month. 

With this, in rslearn prepare, we can specify 1 worker and batch size that larger than number of windows, then it will create much less items than previously and only need to make those requests to CDS API. This save a lot of time in requesting data, thus speed up the ingest process. 

Here's my config.json and commands:

```
{
    "layers": {
        "era5": {
            "type": "raster",
            "band_sets": [{
                "dtype": "float32",
                "bands": ["2t", "tp"]
            }],
            "data_source": {
                "name": "rslearn.data_sources.climate_data_store.ERA5LandMonthlyMeans",
                "query_config": {
                    "max_matches": 3
                }
            }
        }
    }
}
```

```
rslearn dataset add_windows --root $DATASET_PATH --group default --utm --resolution 10 --grid_size 128 --src_crs EPSG:4326 --box=-122.6901,47.2079,-121.4955,47.9403 --start 2024-06-01T00:00:00+00:00 --end 2024-08-01T00:00:00+00:00 --name seattle
rslearn dataset prepare --root $DATASET_PATH --workers 1 --batch-size 5000
rslearn dataset ingest --root $DATASET_PATH --workers 32 --no-use-initial-job --jobs-per-process 1
rslearn dataset materialize --root $DATASET_PATH --workers 32 --no-use-initial-job
```

yawenzzzz:
> The code looks good to me.
> 
> Two additional comments that you can decide whether to consider:
> 
> 1. I only just understood that the item is being constructed based on the requested geometry. In this case, it may be better to implement the data source like https://github.com/allenai/rslearn/blob/master/rslearn/data_sources/xyz_tiles.py where the data is materialized directly from the data source rather than ingesting item and then materializing from the item. This makes more sense for data sources that do not really contain logical items.
> 2. I tested it with this sequence of commands but I couldn't tell whether things were correct, there were some NaN values and such.
> 
> ```
> python -m rslearn.main dataset add_windows --root tmp_dataset/ --group default --utm --resolution 500 --window_size 128 --src_crs EPSG:4326 --box=-122.350558,47.611294,-122.350558,47.611294 --start 2024-06-01T00:00:00+00:00 --end 2024-08-01T00:00:00+00:00 --name seattle
> python -m rslearn.main dataset prepare --root tmp_dataset/ --workers 8
> python -m rslearn.main dataset ingest --root tmp_dataset/
> python -m rslearn.main dataset materialize --root tmp_dataset/ --workers 8
> ```
> 
> ```json
> {
>     "layers": {
>         "sentinel2": {
>             "type": "raster",
>             "band_sets": [{
>                 "dtype": "uint8",
>                 "bands": ["R", "G", "B"]
>             }],
>             "data_source": {
>                 "name": "rslearn.data_sources.gcp_public_data.Sentinel2",
>                 "index_cache_dir": "cache/sentinel2/",
>                 "max_time_delta": "1d",
>                 "sort_by": "cloud_cover",
>                 "use_rtree_index": false
>             }
>         },
>         "era5": {
>             "type": "raster",
>             "band_sets": [{
>                 "dtype": "float32",
>                 "bands": ["2t", "tp"]
>             }],
>             "data_source": {
>                 "name": "rslearn.data_sources.climate_data_store.ERA5LandMonthlyMeans",
>                 "api_key": "8804f983-ac29-4072-98a3-8b04c2cfa8b6"
>             }
>         }
>     }
> }
> ```

Thanks for the suggestion!

For 1, I think it's more true before my last change, when the items are only specific to a single ST geometry, in this case, we can skip the ingestion step and materialize for the window directly. However, my last change is to aggregate the ST geometries to get a union geometry, and use this union geometry to send a request to CDS, in this case, having a tile store is still necessary, since we may end up just ingesting one item, and all windows can use that item since it covers all their geometries. 

For 2, yes, some NaN values are expected, as this is the EAR5 LAND product, and it will give NaN when there's water bodies. There's also an ERA5 product that cover both land and water, but at a even coarser resolution (0.25 degree): https://cds.climate.copernicus.eu/datasets/reanalysis-era5-pressure-levels-monthly-means?tab=overview. When inspecting the NaN value, I found that the ERA5 layer have some 0 values (see black regions below), which is due to the request boundary, therefore, I added some buffer to the request boundaries, to make sure they fully cover the geometries, it's fixed now. 

Screenshot 2024-12-09 at 11 45 05 AM

PR #112, opened 2024-12-09
Add instruction on prepare workers for ERA5

yawenzzzz:


favyen2 (approved):
approved but can you change "preparing" to "ingesting" since preparing in rslearn means a different phase. thanks for adding this doc!

PR #114, opened 2024-12-12
move context setting within main

Hgherzog:
When we call the main cli through rslearn dataset materialize (or a similar command we are calling the main function directly so the mp context is not set from the if __name__ == "__main__":. We set it in the main function instead so it works for the entire rslearn main cli

### DIFF
--- a/rslearn/main.py
+++ b/rslearn/main.py
@@ -712,8 +712,17 @@ def model_predict() -> None:
 
 def main() -> None:
     """CLI entrypoint."""
-    multiprocessing.set_start_method(MULTIPROCESSING_CONTEXT)
-    logger.info(f"Using multiprocessing context: {multiprocessing.get_context()}")
+    try:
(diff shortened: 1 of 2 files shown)


favyen2 (approved):
thanks for fixing, just need to add the try except block (or force)

rslearn/main.py
```diff
@@ -710,6 +712,17 @@ def model_predict() -> None:
...
 def main() -> None:
     """CLI entrypoint."""
+    try:
+        multiprocessing.set_start_method(MULTIPROCESSING_CONTEXT)
+    except RuntimeError as e:
+        logger.error(
+            f"Multiprocessing context already set to {multiprocessing.get_context()}: "
+            + f"ignoring {e}"
+        )
+    except Exception as e:
```
favyen2, line 722: This except seems unnecessary. I would remove it unless issues crop up, otherwise it could hide an unknown problem.

Hgherzog: I should raise here you are right. I forgot to add that

PR #115, opened 2024-12-12
Error handling and ignore errors option for materialize

Hgherzog:


rslearn/main.py
```diff
@@ -590,7 +594,13 @@ def __call__(self, windows: list[Window]) -> None:
         logger.info(f"Running Materialize with {len(windows)} windows")
         if self.dataset is None:
             raise ValueError("dataset not set")
-        materialize_dataset_windows(self.dataset, windows)
+        try:
+            materialize_dataset_windows(self.dataset, windows)
+        except Exception as e:
+            if not self.ignore_errors:
+                logger.error(f"Error materializing windows: {e}")
+                raise e
```
favyen2, line 602: I think `raise` is generally preferred here to `raise e` so that the last part of traceback is retained as the origin of the error rather than appending this line to the traceback.

favyen2 (approved):

PR #118, opened 2024-12-19
Henryh/bumpdockerfiletorch

Hgherzog:
Also, solves dependency issue occuring in rslp where pip picks different version of gcsfs when installing from extra_requirements.txt in pip install rslearn[extra]. This issue appears because there is a new latest gcsfs package available as of this afternoon.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM pytorch/pytorch:2.4.0-cuda11.8-cudnn9-runtime@sha256:58a28ab734f23561aa146fbaf777fb319a953ca1e188832863ed57d510c9f197
+FROM pytorch/pytorch:2.5.0-cuda11.8-cudnn9-runtime@sha256:d15e9803095e462e351f097fb1f5e7cdaa4f5e855d7ff6d6f36ec4c2aa2938ea
 
 # Weirdly this stopped working
 # RUN apt-get update \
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -16,4 +16,4 @@ COPY ./test_requirements.txt /rslearn/test_requirements.txt
 RUN pip install --no-cache-dir -r test_requirements.txt
 
 COPY ./ /rslearn
-RUN pip install --no-cache-dir  .
+RUN pip install --no-cache-dir  rslearn[extra]
```
favyen2, line 19: Won't this install from pypi rather than from the current directory?

What about `pip install --no-cache-dir .[extra]`?

extra_requirements.txt
```diff
@@ -3,7 +3,6 @@ accelerate>=1.0
 cdsapi>=0.7.4
 earthengine-api>=0.1
 einops>=0.8
-gcsfs>=2024.10
```
favyen2, line 6: It looks like s3fs is also covered in the `fsspec[s3,gcs]` so should that be removed from here as well?

Dockerfile
favyen2: Is it needed to change this or is the change to `extra_requirements.txt` sufficient?

Hgherzog: removed this change

favyen2 (approved):
lgtm (although I couldn't understand how the issue would've manifested)

Hgherzog:
> lgtm (although I couldn't understand how the issue would've manifested)
@favyen2 
1. prior to this afternoon 2024.10 was the latest package
2. when we imported in extra requirements from pypi it pulled in the latest which was the same that we resolved too form the normal requirements.
3. Today a new gcsfs package was released so the requirement in extra_requirements and requirements no longer resolved to the same package. 
4. we don't see this issue in rslearn because the requirements are installed sequentially first instead of all at onceas they are from pypi

pbeukema (approved):

PR #126, opened 2025-01-16
Implement data source for ESA WorldCover 2021 data.

favyen2:
Add WorldCover data source. It is a wrapper around LocalFiles that helps with downloading all of the zip files from ESA website.

Also:
* Made some adjustments to LocalFiles to improve performance (use open_rasterio_upath_reader) and not require `config.data_source` to be set when using LocalFiles programmatically.
* Update `GeotiffRasterFormat` to allow setting the filename where raster should be encoded in case it is being used separately from an rslearn dataset (this is used e.g. for populating Helios dataset).


yawenzzzz (approved):
LGTM! Thanks for adding the WorldCover data. Using LocalFile seems like a good approach for static data (large GeoTIFFs that don’t change frequently over time). We can process SRTM, the average DynamicWorld in a very similar way.

PR #131, opened 2025-02-11
Fix bug in AWS Landsat data source where cloud cover is not used correctly

favyen2:
It was string so sorting would not be in the right order.
And then also it wasn't sorting the right list.

I added a test that fails without this fix.


yawenzzzz (approved):
Thanks for the fix!

PR #134, opened 2025-02-13
Sentinel-2 GCP: fix bugs with OPER products and invalid geometry

favyen2:
- Some products have invalid geometry. For example `S2B_MSIL1C_20190111T193659_N0207_R056_T08MLS_20190111T205033` geometry shows up as MultiLineString. These previously could cause problem during prepare (since it is expected that items have positive area not zero area). Now they are skipped (marked corrupt).
- There are products that start with S2A_OPER instead of S2A_MSIL1C (or S2B/C) and these are like weird scenes that shouldn't actually be checked. We were ignoring them before but they could cause error during rtree creation if they have base_url/source_url both invalid (which some do). Now the check is higher up so it avoids the error.

I added test for the first issue but the second issue only comes up when listing all products (with no time range) which takes too long so there doesn't seem to be a good way to test it (the sensing_time of these products seems to be NULL in BigQuery, which is another problem with them).


rslearn/data_sources/gcp_public_data.py
```diff
@@ -269,6 +269,20 @@ def _read_index(
...
+            # products that say OPER instead of MSIL1C.
+            product_id = row["product_id"]
+            product_id_parts = product_id.split("_")
+            if len(product_id_parts) < 7:
+                continue
+            product_type = product_id_parts[1]
+            if product_type != "MSIL1C":
+                continue
+            time_str = product_id_parts[2]
+            tile_id = product_id_parts[5]
```
yawenzzzz, line 283: Suggestion: unpacking list using `_, product_type, time_str, _, _, tile_id, *_ = product_id_parts`

yawenzzzz (approved):
LGTM, only a small suggestion.

rslearn/data_sources/gcp_public_data.py
favyen2: I added some comments with example product IDs to make it more clear what is happening in this code.

I think I prefer the current parsing code over unpacking on one line, even though it's more concise I think it's a bit harder to understand especially with the product_id_parts potentially having different numbers of parts. I tried searching online for some suggestions about which way is preferred but couldn't find much. ChatGPT likes the current one for the clarity but the tuple unpacking for conciseness. Anyway hopefully the comments help to improve the readability.

PR #135, opened 2025-02-14
AWS Landsat: support materializing directly from the COGs

favyen2:
This adds support for skipping ingestion (directly materialize). Similar to the Planetary Computer data sources. User can still ingest if desired (controlled by ingest flag on the data source config).


yawenzzzz (approved):
LGTM.

PR #136, opened 2025-02-14
Consolidate planetary computer data sources

favyen2:
The new generic PlanetaryComputer data source can be used for other collections in their catalog for which we did not make a specific DataSource.


favyen2:
need to update the tests

yawenzzzz (approved):
LGTM.

PR #137, opened 2025-02-14
Add py.typed to pyproject.toml.

### LINKED ISSUE #133: Indicate that rslearn supports type checking
Given that rslearn has type hints, it may make sense to add an empty `rslearn/py.typed` file, which will indicate that the rslearn package supports type checking.

That should help downstream clients who do their own type checking by not requiring them to annotate rslearn imports with `# type: ignore`

Reference:
https://peps.python.org/pep-0561/#packaging-type-information

favyen2:
Resolves #133.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -29,6 +29,9 @@ include = ["rslearn*"]
 dependencies = {file = ["requirements.txt"]}
 optional-dependencies = {extra = { file = ["extra_requirements.txt"] }}
 
+[tool.setuptools.package-data]
+rslearn = ["py.typed"]
+
 [tool.ruff]
 fix = true
 


favyen2:
@APatrickJ can you check if you have time? I'm not sure if I did it correctly but this seems to work:

```
$ python -c "import rslearn, pathlib; print((pathlib.Path(rslearn.__file__).parent / 'py.typed').exists())"
True
```

APatrickJ (approved):
Works for me!

In my code, I temporarily changed my dependency to:

```
rslearn @ git+https://github.com/allenai/rslearn.git@favyen/pytyped
```

And then re-installed the rslearn package.

After that, I removed all of the `# type: ignore` directives from the rslearn imports.

No complaints from `mypy` anymore :-)

Thank you!

PR #139, opened 2025-02-18
Add class_names/class_property_name to dataset config file

favyen2:
These are for annotation / prediction layers to indicate which GeoJSON properties correspond to some notion of a class, or for non-image single-band GeoTIFFs to define what the different values in the band mean.

It could be useful in the future to do things like rasterize a vector layer to a corresponding raster layer; or we can move some similar definitions from model configuration file to dataset configuration file (e.g. make property_name/classes unnecessary in rslearn.train.tasks.classification.ClassificationTask). For now it will be used for ES Studio syncing.


APatrickJ (approved):
This makes sense to me, including the per-band raster settings.

Thanks for doing this so quickly!

PR #140, opened 2025-02-19
Bump Lightning to 2.5.0.post0

favyen2:
This fixes issue with using Python 3.12.8 (incompatibility between Lightning and jsonargparse).

I also bumped fsspec to 2025.2.0. Currently we pin the version because previously pip struggled to install the consistent latest version across fsspec/s3fs/gcsfs.


APatrickJ (approved):
Thanks for fixing this!

PR #143, opened 2025-02-20
Add Copernicus data source.

### LINKED ISSUE #15: Add USGS EROS/EarthExplorer Landsat and ESA Copernicus Sentinel2 implementations.
These will be especially useful for querying for images. Then can combine with `get_item_by_name` from #10.

favyen2:
This retrieves rasters from the Copernicus API. It is currently tested with Sentinel-2 only, and there is a subclass that provides support for Sentinel-2 harmonization, although harmonization may not be needed since ESA has re-processed most if not all of the old scenes.

Resolves #15.

### DIFF
--- a/rslearn/data_sources/copernicus.py
+++ b/rslearn/data_sources/copernicus.py
@@ -3,21 +3,35 @@
 import functools
 import io
 import json
+import os
+import pathlib
 import shutil
+import tempfile
 import urllib.request
 import xml.etree.ElementTree as ET
-import zipfile
 from collections.abc import Callable
+from datetime import datetime
+from enum import Enum
+from typing import Any
+from zipfile import ZipFile
 
 import numpy as np
 import numpy.typing as npt
+import rasterio
+import requests
+import shapely
 from upath import UPath
 
+from rslearn.config import QueryConfig, RasterLayerConfig
 from rslearn.const import WGS84_PROJECTION
+from rslearn.data_sources.data_source import Item
+from rslearn.data_sources.utils import match_candidate_items_to_window
 from rslearn.log_utils import get_logger
+from rslearn.tile_stores import TileStoreWithLayer
 from rslearn.utils.fsspec import open_atomic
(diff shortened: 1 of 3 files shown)


rslearn/data_sources/copernicus.py
```diff
@@ -201,3 +215,632 @@ def get_sentinel2_tiles(geometry: STGeometry, cache_dir: UPath) -> list[str]:
...
+                access_token = self._get_access_token()
+                headers = {
+                    "Authorization": f"Bearer {access_token}",
+                }
+                response = requests.get(
+                    self.DOWNLOAD_URL + path,
+                    stream=True,
+                    headers=headers,
+                    timeout=self.timeout,
+                )
```
hunterp, line 634: Maybe they fixed this; but we saw this issue with Copernicus that they would actually redirect us for the actual download over to some other host; but it still required the auth header. Which the requests library doesn't do by default. 

This is what we did to handle it: 
https://github.com/VulcanSkylight/skylight-sat-service/blob/395f8f112b22bfcfe636c3673404dc06c60dd8c9/sat-py/skylight/sat/downloader/sentinel_dataspace_api_client.py#L129-L146

But again, that was 1.5 years ago, they might have fixed that since.

hunterp (approved):

rslearn/data_sources/copernicus.py
favyen2: Ah yeah I did encounter that issue too, I fixed it with this:

```
    # We use a different URL for downloads. The BASE_URL would redirect to this URL but
    # it makes it difficult since requests library drops the Authorization header after
    # the redirect. So it is easier to access the DOWNLOAD_URL directly.
    DOWNLOAD_URL = "https://download.dataspace.copernicus.eu/odata/v1"
```

https://github.com/allenai/rslearn/blob/a0394ebb908253e2dee89b7e73f5163c0bbf5a2d/rslearn/data_sources/earthdata_srtm.py#L241 has similar issue where the solution there is closer to the code you linked.

PR #144, opened 2025-02-25
Add data source for SRTM elevation data from NASA Earthdata.

### LINKED ISSUE #102: Add SRTM data source
This is a digital elevation map created by NASA's Shuttle Radar Topography Mission.

We need to make sure we're using the V3 product (SRTM Plus) which has:

* 30m / 1 arc-second spatial resolution
* Void filling (see https://lpdaac.usgs.gov/documents/13/SRTM_Quick_Guide.pdf)

USGS provides small scale downloads via EarthExplorer.

For a bulk download, this looks like one possible source:
https://portal.opentopography.org/raster?opentopoID=OTSRTM.082015.4326.1

favyen2:
Add data source for SRTM elevation data from NASA Earthdata.

Resolves #102.

I also added some missing documentation and adjusted two print statements to be debug-only.

### DIFF
--- a/rslearn/data_sources/earthdata_srtm.py
+++ b/rslearn/data_sources/earthdata_srtm.py
@@ -78,6 +78,8 @@ class SRTM(DataSource):
         if config.data_source is None:
             raise ValueError("config.data_source is required")
         d = config.data_source.config_dict
+        if len(config.band_sets) != 1:
+            raise ValueError("expected a single band set")
         if len(config.band_sets[0].bands) != 1:
             raise ValueError("expected band set to have a single band")
         kwargs: dict[str, Any] = {
--- a/rslearn/data_sources/earthdata_srtm.py
+++ b/rslearn/data_sources/earthdata_srtm.py
@@ -47,7 +47,7 @@ class SRTM(DataSource):
         username: str | None = None,
         password: str | None = None,
         band_name: str = "srtm",
-        timeout: int = 10,
+        timeout_seconds: float = 10,
     ):
         """Initialize a new SRTM instance.
 
@@ -57,10 +57,10 @@ class SRTM(DataSource):
             password: NASA Earthdata account password. If not set, it is read from the
                 NASA_EARTHDATA_PASSWORD environment variable.
             band_name: what to call the band.
-            timeout: timeout for downloading data.
+            timeout_seconds: timeout for downloading data.
         """
         self.band_name = band_name
-        self.timeout = timeout
+        self.timeout_seconds = timeout_seconds
 
         if username is None:
             username = os.environ["NASA_EARTHDATA_USERNAME"]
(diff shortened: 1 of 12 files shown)


rslearn/data_sources/earthdata_srtm.py
```diff
@@ -0,0 +1,284 @@
...
+        Args:
+            username: NASA Earthdata account username. If not set, it is read from the
+                NASA_EARTHDATA_USERNAME environment variable.
+            password: NASA Earthdata account password. If not set, it is read from the
+                NASA_EARTHDATA_PASSWORD environment variable.
+            band_name: what to call the band.
+            timeout: timeout for downloading data.
+        """
+        self.band_name = band_name
+        self.timeout = timeout
```
APatrickJ, line 63: Suggest either using `timedelta` for this (preferred) or naming it `timeout_seconds` / `timeout_secs` to indicate the units

rslearn/data_sources/earthdata_srtm.py
```diff
@@ -0,0 +1,284 @@
...
+
+        self.session = requests.session()
+
+    @staticmethod
+    def from_config(config: RasterLayerConfig, ds_path: UPath) -> "SRTM":
+        """Creates a new SRTM instance from a configuration dictionary."""
+        if config.data_source is None:
+            raise ValueError("config.data_source is required")
+        d = config.data_source.config_dict
+        if len(config.band_sets[0].bands) != 1:
```
APatrickJ, line 81: Do we always expect to have a single band set too? Should that be checked as well?

APatrickJ (commented):
Makes sense to me. Thanks for doing this! A couple of minor questions / suggestions.

rslearn/data_sources/earthdata_srtm.py
favyen2: I changed it to timeout_seconds. I think we mostly have this just because of bandit rule that discourages using requests without timeout; I guess it could be useful if it is hanging for hours. I used float instead of timedelta to match the timeout parameter of https://requests.readthedocs.io/en/latest/api/#requests.request, what do you think?

rslearn/data_sources/earthdata_srtm.py
favyen2: Thanks, I think you're right, I thought maybe there was a use case for have multiple band sets with different zoom_offsets or storage formats or something like that but I realized currently that wouldn't work anyway (since they would share the same raster directory because that is based on the band name(s) only).

favyen2:
welp I guess the website goes on and off sometimes (can't access https://e4ftl01.cr.usgs.gov/ today) ._.

or maybe they ran out of money

rslearn/data_sources/earthdata_srtm.py
APatrickJ: You can always convert a `timedelta` value to seconds using `.total_seconds()` when you need to, e.g. when passing it to the requests library. That method returns a float. I find `timedelta` to be a clean way to encode time durations without having to remember what the units are or encode the units as part of the variable name. But `timeout_seconds` does the job too.

APatrickJ (approved):
Looks good (assuming the integration tests can ever pass again!)

PR #145, opened 2025-02-26
Fix CDS API issue

yawenzzzz:
Change band names to the "Variable name in CDS": https://confluence.ecmwf.int/display/CKB/ERA5-Land%3A+data+documentation

### DIFF
--- a/rslearn/data_sources/climate_data_store.py
+++ b/rslearn/data_sources/climate_data_store.py
@@ -45,8 +45,8 @@ class ERA5LandMonthlyMeans(DataSource):
     DOWNLOAD_FORMAT = "unarchived"
     PIXEL_SIZE = 0.1  # degrees, native resolution is 9km
 
-    # see: https://confluence.ecmwf.int/display/CKB/ERA5-Land%3A+data+documentation
-    # for variable names and descriptions
+    # CDS variable names reference: https://confluence.ecmwf.int/display/CKB/ERA5-Land%3A+data+documentation
+    # Replace "_" with "-" in variable names when specifying bands in the data configuration
 
     def __init__(
         self,
--- a/rslearn/data_sources/climate_data_store.py
+++ b/rslearn/data_sources/climate_data_store.py
@@ -46,7 +46,7 @@ class ERA5LandMonthlyMeans(DataSource):
     PIXEL_SIZE = 0.1  # degrees, native resolution is 9km
 
     # see: https://confluence.ecmwf.int/display/CKB/ERA5-Land%3A+data+documentation
-    # For variable full & short names
+    # Use "Variable name in CDS" for the API request
 
     def __init__(
         self,
--- a/rslearn/data_sources/climate_data_store.py
+++ b/rslearn/data_sources/climate_data_store.py
@@ -235,9 +235,10 @@ class ERA5LandMonthlyMeans(DataSource):
                 if band in bands:
                     continue
                 bands.append(band)
-
+        # rslearn band names, remove "_"
+        band_names = [band.replace("_", " ") for band in bands]
         for item in items:
-            if tile_store.is_raster_ready(item.name, bands):
+            if tile_store.is_raster_ready(item.name, band_names):
                 continue
             # Send the request to the CDS API
             # Given that ERA5 is at a very coarse resolution, we add a buffer of half a pixel size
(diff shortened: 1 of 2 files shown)


rslearn/data_sources/climate_data_store.py
```diff
@@ -235,9 +235,10 @@ def ingest(
                 if band in bands:
                     continue
                 bands.append(band)
-
+        # rslearn band names, remove "_"
+        band_names = [band.replace("_", " ") for band in bands]
```
favyen2, line 239: I think the band names in the BandSet configuration should correspond to what is used to construct the directory name (e.g. in call to `tile_store.is_raster_ready`), otherwise the materialize step and other code may fail (they will use the band names from the BandSet and end up with the incorrect directory name).

So I would either:

1. Use spaces for the band names in the BandSet configuration (e.g. in TEST_BANDS) and then convert from " " -> "_" (instead of  "_" -> " ") when getting the names to use for the API query. This way it is always spaces from the perspective of all code outside of `climate_data_store.py`. I would suggest using hyphens instead of spaces though.
2. Use underscores in the band names but change the code that encodes the bands to escape the underscore somehow.

favyen2 (commented):
I think materialization will not work currently due to mismatch in the band names.

rslearn/data_sources/climate_data_store.py
yawenzzzz: Ah I see! Approach 1 sounds good to me, the CDS variable names only include "_" and words, so it's safe to replace "_" to "-" as band names, then convert back to variable names for API request.

tests/integration/data_sources/test_climate_data_store.py
```diff
@@ -21,10 +21,12 @@
 class TestERA5LandMonthlyMeans:
     """Tests the ERA5LandMonthlyMeans data source from the Climate Data Store."""
 
-    TEST_BANDS = ["2t", "tp"]  # 2m temperature and total precipitation
+    TEST_BANDS = ["2m-temperature", "total-precipitation"]
 
     def run_simple_test(self, tile_store_dir: UPath, seattle2020: STGeometry) -> None:
         """Apply test where we ingest an item corresponding to seattle2020."""
+        # rslearn band names, remove "_"
+        band_names = [band.replace("_", " ") for band in self.TEST_BANDS]
```
favyen2, line 29: no longer needed

tests/integration/data_sources/test_climate_data_store.py
```diff
@@ -49,8 +51,8 @@ def run_simple_test(self, tile_store_dir: UPath, seattle2020: STGeometry) -> Non
         data_source.ingest(
             TileStoreWithLayer(tile_store, layer_name), item_groups[1], [[seattle2020]]
         )
-        assert tile_store.is_raster_ready(layer_name, item_0.name, self.TEST_BANDS)
-        assert tile_store.is_raster_ready(layer_name, item_1.name, self.TEST_BANDS)
+        assert tile_store.is_raster_ready(layer_name, item_0.name, band_names)
```
favyen2, line 54: should be able to use self.TEST_BANDS here

favyen2 (approved):
Sounds good but I think the `band_names` is no longer needed in the test.

tests/integration/data_sources/test_climate_data_store.py
```diff
@@ -21,10 +21,11 @@
 class TestERA5LandMonthlyMeans:
     """Tests the ERA5LandMonthlyMeans data source from the Climate Data Store."""
 
-    TEST_BANDS = ["2t", "tp"]  # 2m temperature and total precipitation
+    TEST_BANDS = ["2m-temperature", "total-precipitation"]
 
     def run_simple_test(self, tile_store_dir: UPath, seattle2020: STGeometry) -> None:
         """Apply test where we ingest an item corresponding to seattle2020."""
+        # rslearn band names, remove "_"
```
favyen2, line 28: Can you remove this comment as well? And in `climate_data_store.py` it would likely be useful to add a note about how to get the variable name and that user should replace underscore with hyphen.

tests/integration/data_sources/test_climate_data_store.py
yawenzzzz: Ah right, forgot to update it as tests passed...

tests/integration/data_sources/test_climate_data_store.py
yawenzzzz: Fixed, also add note about the variable names.

PR #146, opened 2025-03-03
Improve the API for RasterFormat and VectorFormat.

favyen2:
Improve the API for RasterFormat and VectorFormat.

RasterFormat: the previous API accepts a projection and bounds when encoding, but only bounds when decoding. This commit changes it to also accept projection when decoding, and the RasterFormat is expected to re-project the data to the requested projection. Previously, the caller had to do extra work to ensure the bounds provided exactly matches the projection of the actual raster. Now, if the caller wants data in a different projection they can pass it rather than re-projecting themselves.

VectorFormat: the previous API accepts a projection when encoding (no bounds), and bounds when decoding (no projection). Encoding is changed to accept neither projection nor bounds: neither are necessary since the Features include a projection. Decoding is changed to accept both a projection and bounds for similar reasons to RasterFormat.

The original motivation of this change was because of issues employing TileVectorFormat for the vector format used in DefaultTileStore, since the API didn't really support a way for windows in different projections to ask for portions of the vector data.

I think the new API makes a lot more sense because (a) passing around bounds without a corresponding projection is strange especially when the formats don't necessarily have a way to get the projection of the stored data, and (b) there was already some re-projecting logic in DefaultTileStore which now is just moved to VectorFormat/RasterFormat.

### DIFF
--- a/rslearn/config/dataset.py
+++ b/rslearn/config/dataset.py
@@ -190,8 +190,8 @@ class BandSetConfig:
         return BandSetConfig(**kwargs)  # type: ignore
 
     def get_final_projection_and_bounds(
-        self, projection: Projection, bounds: PixelBounds | None
-    ) -> tuple[Projection, PixelBounds | None]:
+        self, projection: Projection, bounds: PixelBounds
+    ) -> tuple[Projection, PixelBounds]:
         """Gets the final projection/bounds based on band set config.
 
         The band set config may apply a non-zero zoom offset that modifies the window's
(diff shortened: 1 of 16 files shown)


favyen2:
This PR touches a lot of code but the main changes are in `rslearn/utils/raster_format.py` and `rslearn/utils/vector_format.py` to change the API there.

I also did some other cleanup/changes:
- Remove None for type option when we aren't using it.
- Remove the zoom_offset option from VectorLayerConfig. I think I added this originally because I thought we might store vector data as integer and then it might be useful to get higher precision for some layers in the same way we support zoom_offset for RasterLayerConfig, but since it is floating point this doesn't actually seem to make sense.
- Support customizing the vector format in DefaultTileStore (this was the original motivation for the PR, because I wanted to use TileVectorFormat so that I could materialize OpenStreetMap data without the materialization workers needing to load the entire GeoJSON file every time).

APatrickJ (approved):
The changes make sense to me! Overall it looks like a very good cleanup.

PR #148, opened 2025-03-03
Reduce the STAC calls made by PlanetaryComputer during ingestion/materialization

favyen2:
Reduce the STAC calls made by PlanetaryComputer during ingestion/materialization.

During direct materialization in particular, we previously would make 3 STAC calls to get STAC item by ID: once in get_raster_bands, once in get_raster_bounds, and once in read_raster.

Now we expose a cache_dir option so we save the asset URLs during preparing, so we don't need any of these calls. Although the asset URLs are now saved with the item, the cache_dir is still needed because we don't currently have access to the item objects when materializing with a TileStore interface.

This speeds up direct materialization substantially since previously the STAC API calls were a bottleneck.


yawenzzzz (approved):
LGTM!

PR #149, opened 2025-03-03
Inject custom serializer/deserializer for rasterio.crs.CRS.

favyen2:
Inject custom serializer/deserializer for rasterio.crs.CRS.

This is needed because rasterio does not support typing. It enables configuring a DefaultTileStore with a TileVectorFormat that has a projection set.


yawenzzzz (approved):
LGTM!

PR #151, opened 2025-03-07
Updates needed for Sentinel-2 vessel attribute prediction.

favyen2:
Updates needed for Sentinel-2 vessel attribute prediction.

- Add process_output to RegressionTask
- Add a new metric for RegressionTask that checks if the output is within a multiplicative factor of the ground truth value.
- Also some miscellaneous cleanup, including removing unused open_atomic in rslearn/utils/utils.py (has been replaced with open_atomic in rslearn/utils/fsspec.py) and fixing type of process_output in Task base class.


rslearn/train/tasks/regression.py
```diff
@@ -241,3 +283,46 @@ def reset(self) -> None:
     def plot(self, *args: list[Any], **kwargs: dict[str, Any]) -> Any:
         """Returns a plot of the metric."""
         return self.metric.plot(*args, **kwargs)
+
+
+class WithinFactorMetric(Metric):
```
yawenzzzz, line 288: Calling this `Accuracy` for regression task could be more intuitive, the within factor adds tolerance into prediction and then we compute the % of correctness similarly as classification task.

yawenzzzz (commented):
LGTM! Only a minor comment.

rslearn/train/tasks/regression.py
favyen2: Thanks, that makes sense, renamed and added a test.

yawenzzzz (approved):

PR #154, opened 2025-03-26
Update Sentinel1 and Sentinel2 to pass timedelta

### LINKED ISSUE #153: Update Sentinel1 and Sentinel2 PlanetaryComputer implementations to pass timedelta
PlanetaryComputer was recently changed to accept a `timeout: timedelta` parameter instead of `timeout_seconds: float`. The subclasses need to be updated to pass the correct parameter.

APatrickJ:
Fixes #153

### DIFF
--- a/rslearn/data_sources/planetary_computer.py
+++ b/rslearn/data_sources/planetary_computer.py
@@ -551,16 +551,13 @@ class Sentinel2(PlanetaryComputer):
             assets=list(needed_assets),
         )
 
+        if "timeout_seconds" in d:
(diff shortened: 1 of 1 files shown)


favyen2 (approved):

favyen2:
thanks for the fix!

APatrickJ:
> I guess we will need tests for the config parsing but hoping to change that eventually so maybe not worth testing for now.

Agreed on both. I thought about adding tests, but as I didn't see any existing ones for config parsing I wasn't sure if there's a particular way you want to model them or a reason they were left out. I can take a look at that in a followup if you'd like.

PR #155, opened 2025-04-03
Improvements to model components.

favyen2:
- Conv: support setting padding as integer which is needed if using stride.
- Conv: fix forward pass type to Any since it is ignored.
- module_wrapper: add EncoderModuleWrapper to wrap a decoder component like Conv to be used in the encoder.


yawenzzzz (approved):
LGTM!

PR #156, opened 2025-04-03
Fix bug with multiple tags in a SplitConfig.

favyen2:
Previously if there are multiple tags, then a window would be added as a training example for each tag that it matches. This means there would be duplicates, and also it doesn't match the intended behavior where the tags are additive constraints (i.e. the overall op should be logical and of all the specified tags).

This commit fixes the bug.


yawenzzzz (approved):
LGTM!

PR #157, opened 2025-04-03
Improve ERA5 data source to work with larger-scale window collections.

favyen2:
Previously if the windows have global coverage, the data source would make a request for data across the world, but this request would fail for some reason.

Additionally, it is like OpenStreetMap data source where the items are dependent on the requested windows rather than independent.

This changes ERA5 to represent the underlying data as fixed 10x10 degree items (even though the API supports arbitrary geometry requests) and matches those to windows. The size of an item can be configured if desired but the point is there is a fixed grid. Temporally it is still gridded by calendar month.


yawenzzzz (approved):
LGTM! Thanks for the changes!

PR #162, opened 2025-04-08
Improve mosaic creation by adding items to ladder of mosaics.

### LINKED ISSUE #159: Improve mosaic creation for max_matches>1
Currently we create mosaics one at a time. So we process the items in order and discard items that do not contribute new regions to the mosaic. If there is a missing portion of the window, then only one partial mosaic will be created (rather than multiple partial mosaics). If all the items covering one portion of the window appear before items covering another portion, then only one mosaic may be created that covers that first portion.

We should instead process the items in order but use the duplicative items to create additional mosaics instead of discarding them.

favyen2:
Previously we would iterate over items, and if an item didn't add coverage to the current mosaic then we would skip it. Now we push it down to the next mosaic and so on to see if it can be added to any of them.

Resolves #159.

### DIFF
--- a/rslearn/data_sources/aws_landsat.py
+++ b/rslearn/data_sources/aws_landsat.py
@@ -109,7 +109,6 @@ class LandsatOliTirs(DataSource, TileStore):
         self.config = config
         self.metadata_cache_dir = metadata_cache_dir
         self.sort_by = sort_by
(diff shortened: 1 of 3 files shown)


APatrickJ (approved):
Thanks for explaining this! 👍

PR #166, opened 2025-04-25
Improve Planetary Computer error handling.

favyen2:
Improve Planetary Computer error handling.

Sometimes prepare fails while initializing the worker processes because when they initialize the PlanetaryComputer data source it fails to create the STAC client. Now it is loaded lazily so it can be caught by retry loop or ignore errors parameter.


APatrickJ (approved):
Very nice change!

PR #168, opened 2025-05-07
Fix bug when tags aren't specified on the window

favyen2:
https://github.com/allenai/rslearn/pull/156 introduced a bug where, if a window doesn't have a tag that's set in the model configuration file, then it will cause an error (KeyError). This PR fixes that bug while also adding a test.


yawenzzzz (approved):
LGTM!

PR #169, opened 2025-05-07
Add CROMA model

favyen2:


rslearn/models/croma.py
```diff
@@ -0,0 +1,230 @@
...
+DEFAULT_IMAGE_RESOLUTION = 120
+PRETRAINED_URLS: dict[CromaSize, str] = {
+    CromaSize.BASE: "https://huggingface.co/antofuller/CROMA/resolve/main/CROMA_base.pt",
+    CromaSize.LARGE: "https://huggingface.co/antofuller/CROMA/resolve/main/CROMA_large.pt",
+}
+MEAN_AND_STD_BY_MODALITY: dict[str, tuple[float, float]] = {
+    "sentinel1": (-15, 10),
+    "sentinel2": (1500, 1500),
+}
+
```
yawenzzzz, line 47: I agree that a small batch size may not give reliable statistics. But this approach also ignores the varying mean and std across Sentinel2 bands. I wonder if it’s better to use batch statistics or the Helios normalization stats instead.

yawenzzzz (approved):
LGTM! Just a minor comment about normalization.

rslearn/models/croma.py
favyen2: Initially I copied CROMA implementation with batch statistics but it did not seem to perform well, I think it diverged during training and didn't handle uniform images (like completely cloudy) well. If normalization from Helios dataset stats performs better then we can switch to that.

favyen2: OK I am running some experiments with per-band stats

favyen2: It seems to have similar performance, but closer to the "right statistics", so will push the change.

PR #170, opened 2025-05-07
Make unet and simple_time_series useful for Helios model.

favyen2:
For unet decoder, since Helios outputs a single feature map with a large embedding depth, we need to control the number of features as it upsamples them. So new optional parameter num_channels dict is added to configure this.

For simple_time_series, I added image_key parameter so we can apply the time series on an image input that is not named "image". This way it can be used e.g. for "sentinel2_l2a". Currently it still only supports one image_key so it can't be used for multimodal input.


yawenzzzz (approved):
LGTM!

PR #176, opened 2025-05-13
Add DETR object detection head.

favyen2:
I haven't been able to train the head successfully but it seems to be correct? I guess I should try it on COCO.

Most of the code is from here:
https://github.com/facebookresearch/detr

I wasn't able to use it as a library because their repository is not structured as a module. Also it was too tightly connected to the backbone.


favyen2:
Henry suggested trying it with other backbone (not Helios).

favyen2:
I should also try longer training run with Helios+DETR with dropout enabled, most recently I tried disabling it but maybe it works better enabled (previously I was seeing very different train/eval results so that's why I disabled it).

PR #178, opened 2025-06-03
Add data sources for non-terrain-corrected Sentinel-1 scenes

favyen2:


rslearn/tile_stores/default.py
```diff
@@ -268,15 +273,40 @@ def write_raster_file(
...
+                # the raster might actually need warping.
+                if profile["crs"] is None and src.gcps:
+                    gcps, gcp_crs = src.gcps
+                    # Use the first ground control point to pick a UTM/UPS projection.
+                    first_gcp_orig = STGeometry(
+                        Projection(gcp_crs, 1, 1),
+                        shapely.Point(gcps[0].x, gcps[0].y),
+                        None,
+                    )
+                    first_gcp_wgs84 = first_gcp_orig.to_projection(WGS84_PROJECTION)
```
yawenzzzz, line 289: This looks reasonable to me, but what if the gcps span a large area, if that's the case, will getting the centroid (ie average) of gcps be a better choice?

yawenzzzz (approved):
LGTM! Just a minor comment regarding the gcps.

rslearn/tile_stores/default.py
favyen2: I think I was worried if the GeoTIFF spans the anti-meridian and then it is more complex to handle figuring out what the centroid is.

favyen2: I looked at it again and yeah the averaging across anti-meridian code seems like it will be overly complicated.

PR #188, opened 2025-06-24
Support nodata values other than zero through config

favyen2:
Support nodata values other than zero through config

### DIFF
--- a/rslearn/dataset/materialize.py
+++ b/rslearn/dataset/materialize.py
@@ -60,6 +60,7 @@ def read_raster_window_from_tiles(
     bounds: PixelBounds,
     src_indexes: list[int],
     dst_indexes: list[int],
+    nodata_vals: list[float],
     remapper: Remapper | None = None,
     resampling: Resampling = Resampling.bilinear,
 ) -> None:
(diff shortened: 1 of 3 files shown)


rslearn/dataset/materialize.py
```diff
@@ -152,10 +160,18 @@ def materialize(
             )
 
             for group_id, group in enumerate(item_groups):
+                # Initialize the destination array to the nodata values.
+                # We default the nodata value to 0.
+                nodata_vals = band_cfg.nodata_vals
+                if nodata_vals is None:
+                    nodata_vals = [0 for _ in band_cfg.bands]
```
yawenzzzz, line 167: Question: Why default nodata value and dst to 0? 0 can be normal value for some of the bands, for example, total precipitation, default to specialy value like -9999 maybe better?

yawenzzzz (approved):
LGTM! Just a minor comment about the default value for NoData

rslearn/dataset/materialize.py
favyen2: I think this is mostly because that was the previous implementation so ideally the change would be backwards compatible.

PR #194, opened 2025-07-03
Support matching per-period mosaics and add min_matches option.

### LINKED ISSUE #160: Move monthly data sources to rslearn
In rslearn_projects, we implemented a few data sources that wrap existing data sources but change the matching functionality to create per-interval mosaics (instead of creating the mosaics by iterating through the items in their provided order).

https://github.com/allenai/rslearn_projects/blob/master/rslp/satlas/data_sources.py

It seems like these per-interval mosaics will be a common use case for applications operating over a long time range. So we should move it to rslearn (in the QueryConfiguration).

favyen2:
The per-period mosaic code is copied from rslearn_projects and can be removed from there after this is merged.

Resolves #160.


rslearn/data_sources/utils.py
```diff
@@ -37,6 +37,163 @@ class PendingMosaic:
...
+            "all windows must have time range for per period mosaic matching"
+        )
+
+    # For each period, we create an STGeometry with modified time range matching that
+    # period, and use it with match_candidate_items_to_window to get a mosaic.
+    cur_groups: list[list[ItemType]] = []
+    period_end = window_geometry.time_range[1]
+    while period_end > window_geometry.time_range[0] and len(cur_groups) < max_matches:
+        period_time_range = (
+            period_end - period_duration,
```
APatrickJ, line 171: For my own understanding, what is the reason for not using `max(period_end - period_duration, window_geometry.time_range[0])`

APatrickJ (approved):
This looks good, I have just one question but it isn't blocking.

rslearn/data_sources/utils.py
favyen2: I don't think there is a specific reason beyond needing to make some choice. It would also make sense to do `while period_end - period_duration > window_geometry.time_range[0]` instead of `while period_end > window_geometry.time_range[0]`; that may be better so that the periods are always the same duration, while also being contained within the window time range. I'm not sure it is important to keep in the window time range though, e.g. if time range is two weeks but period is 30 days it may still make sense to create one mosaic.

APatrickJ: I'm okay either way, but would prefer it's documented explicitly. My thought is that having at least one mosaic is likely to be the more desirable outcome most of the time, but that isn't based on any evidence just intuition.

Is this something that would make sense to adjust based on `time_mode`? When we're collecting constituent images that collectively make up a mosaic, could the set of source imagery be chosen based on what matches for the selected time mode?

favyen2: Changed to limit it to the window duration and documented it in the docstring.

APatrickJ (approved):

PR #195, opened 2025-07-03
support reweighting how class is selected from probabilities in segmentationtask

favyen2:
support reweighting how class is selected from probabilities in segmentationtask


Hgherzog (approved):
LGTM

PR #196, opened 2025-07-03
Add WorldCereal confidences

gabrieltseng:


rslearn/data_sources/local_files.py
```diff
@@ -93,8 +93,15 @@ def from_config(src_dir: UPath, d: dict[str, Any]) -> "RasterItemSpec":
...
+
+        def fname_from_src_suffix(src_dir: UPath, suffix: str | UPath) -> UPath:
+            if isinstance(suffix, UPath):
+                return suffix
+            else:
+                return join_upath(src_dir, suffix)
+
         kwargs = dict(
-            fnames=[join_upath(src_dir, suffix) for suffix in d["fnames"]],
+            fnames=[fname_from_src_suffix(src_dir, suffix) for suffix in d["fnames"]],
```
gabrieltseng, line 104: This lets us pass a suffix which is the full filepath of the file (which is helpful because I use `glob` to obtain it in the worldcereal data source).

tests/integration/conftest.py
```diff
@@ -14,18 +13,18 @@
...
-    test_bucket = os.environ["TEST_BUCKET"]
-    print(f"test_bucket: {test_bucket}")
-    storage_client = storage.Client()
-    try:
-        storage_client.get_bucket(test_bucket)
-        print(f"Bucket {test_bucket} exists.")
-    except Exception as e:
-        raise AssertionError(f"Bucket {test_bucket} does not exist: {str(e)}")
-    return test_bucket
+# @pytest.fixture(scope="session", autouse=True)
```
gabrieltseng, line 16: I had to comment it out for my local tests so not autousing would be nice !

rslearn/data_sources/local_files.py
favyen2: Does it work to pass str suffix to join_upath derived from a `full_path: UPath` using `full_path.absolute().as_uri()`? That should yield protocol at beginning of the returned string which should cause `join_upath` to treat it as absolute path.

rslearn/data_sources/worldcereal.py
```diff
@@ -0,0 +1,230 @@
...
+from rslearn.log_utils import get_logger
+from rslearn.utils.fsspec import get_upath_local, join_upath, open_atomic
+
+logger = get_logger(__name__)
+
+
+class WorldCerealConfidences(LocalFiles):
+    """A data source for the ESA WorldCereal 2021 agricultural land cover map.
+
+    For details about the land cover map, see https://esa-worldcereal.org/en.\
```
favyen2, line 24: extra backslash I think

rslearn/data_sources/worldcereal.py
```diff
@@ -0,0 +1,230 @@
...
+            spec_dict = {
+                # must be a str since we / with a posix path later
+                "name": str(aez),
+                "fnames": [],
+                "bands": [],
+            }
+            for band, tif_path in tif_filepaths.items():
+                aez_band_filepath = self.filepath_for_product_aez(tif_path, aez)
+                if aez_band_filepath is not None:
+                    cast(list, spec_dict["fnames"]).append(aez_band_filepath)
```
favyen2, line 78: Is this just for type checker? If so, and if it works to instead declare `spec_dict: dict[str, Any] = { ...` then I think that is cleaner.

rslearn/utils/raster_format.py
```diff
@@ -423,6 +423,8 @@ def encode_raster(
         """
         if fname is None:
             fname = self.fname
+        else:
+            self.fname = fname
```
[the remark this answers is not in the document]
favyen2: The `self.fname` isn't capitalized but it was intended to be more like a constant. And then the `fname` passed to function can override it in case the caller is operating outside of the normal rslearn layer directory and wants to read something that has a different name.

favyen2 (approved):
Looks good to me! Left a few minor comments but can be merged if the overall materialization with actual data has been verified to work.

tests/integration/conftest.py
favyen2: This function seems to verify that the TEST_BUCKET exists but doesn't seem to really be needed. So I think it could be removed if it is causing problems. The TEST_BUCKET should already be set in the Github action so the default isn't strictly needed.

rslearn/utils/raster_format.py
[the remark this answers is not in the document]
gabrieltseng: okay - i think i changed it because below in `get_raster_bounds`, it calls `self.fname`:

```python
    def get_raster_bounds(self, path: UPath) -> PixelBounds:
        """Returns the bounds of the stored raster.

        Args:
            path: the directory where the raster data was written

        Returns:
            the PixelBounds of the raster
        """
        with open_rasterio_upath_reader(path / self.fname) as src:
            _, bounds = get_raster_projection_and_bounds(src)
            return bounds
```

Would it make sense to create a new attribute (`default_fname` and `fname`) to handle this case?

tests/integration/conftest.py
gabrieltseng: okay. I will leave it for now since its no longer causing me issues

rslearn/utils/raster_format.py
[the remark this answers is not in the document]
gabrieltseng: yes you are right - I misremembered why I changed it. Updated

PR #199, opened 2025-07-08
Fix freeze/unfreeze on restore

ryspark:
Fixes small bug where `FreezeUnfreeze` freezes models upon restoring a run, but never unfreezes them. Note also unfreezing causes even the BatchNorm statistics to get recomputed, not sure if this is intended behavior.


favyen2 (approved):
thanks for fixing this!

PR #201, opened 2025-07-08
Improve the configuration of optimizer and learning rate scheduler

### LINKED ISSUE #186: Optimizer and scheduler configuration
It should be more flexible with directly passing in the optimizer / scheduler (or factory for it) as argument.

favyen2:
Improve the configuration of optimizer and learning rate scheduler

Resolves #186

### DIFF
--- a/rslearn/train/callbacks/freeze_unfreeze.py
+++ b/rslearn/train/callbacks/freeze_unfreeze.py
@@ -3,7 +3,12 @@
 import torch
 from lightning.pytorch import LightningModule
 from lightning.pytorch.callbacks import BaseFinetuning
+from torch.optim.lr_scheduler import ReduceLROnPlateau
 from torch.optim.optimizer import Optimizer
+
+from rslearn.log_utils import get_logger
+
+logger = get_logger(__name__)
 
 
 class FreezeUnfreeze(BaseFinetuning):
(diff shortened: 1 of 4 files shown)


rslearn/train/callbacks/freeze_unfreeze.py
```diff
@@ -70,10 +70,10 @@ def finetune_function(
                 optimizer=optimizer,
                 initial_denom_lr=self.unfreeze_lr_factor,
             )
-            if "plateau" in pl_module.schedulers:
-                scheduler = pl_module.schedulers["plateau"]
+            if "scheduler" in pl_module.schedulers:
+                scheduler = pl_module.schedulers["scheduler"]
                 while len(scheduler.min_lrs) < len(optimizer.param_groups):
-                    print("appending to plateau scheduler min_lrs")
+                    print("appending to scheduler min_lrs")
```
ryspark, line 76: Isn't `min_lrs` specific to the plateau scheduler? Don't know if other schedulers require param-group-specific attributes too

favyen2: Ah okay thanks, I updated the code back to check it but with `isinstance`

favyen2: if there are more special cases then probably can see if it can be moved to the SchedulerFactory

ryspark (approved):

PR #203, opened 2025-07-08
Fix model tests by skipping large model test and deleting cached checkpoints

favyen2:


yawenzzzz (approved):

ryspark (approved):

favyen2:
> Doesn't deleting the cache in `test_terramind.py` delete all downloaded HF models, even unrelated ones? I don't know if anyone runs the tests locally though, or if it's just Github Actions?

Good point, thanks! Fixed it to only do the deletion in CI.

PR #204, opened 2025-07-10
Supporting building and saving an index over windows to speed up window loading

### LINKED ISSUE #161: Support window index
Support creating an index over the windows containing the metadata.

This can just be a JSON file, it is not that large but it avoids needing to read potentially thousands of objects.

It should include both metadata.json and items.json for all the windows.

favyen2:
Resolves #161.

Currently, loading windows can take several minutes or even hours, depending on the number of windows and where the dataset is stored. This is especially annoying when we need to wait for the loading to complete before the model starts training.

This PR adds a DatasetIndex. The index is stored in a single file/object, so it can be retrieved relatively quickly; it isn't compressed, so we read about the same number of bytes, but reading a single file is much faster than reading hundreds of thousands of files. The index enumerates the windows and their metadata (metadata.json), along with the layer datas (items.json) and which layers are "completed".

A new `rslearn dataset build_index` command is added, and we check if the index exists when calling Dataset.load_windows. If it does exist, then the index is passed to the returned Window objects so that they can continue to use the index to satisfy calls to load_layer_datas / is_layer_completed.

Additionally, I updated `rslearn.train.dataset.ModelDataset` to save the list of Windows to a file rather than keeping it as a field. The list is then loaded from the file lazily after the dataloader workers are started. Keeping it as a field means pytorch sends it to each worker when it spawns it sequentially, which is much slower. This further speeds up the initialization time for training jobs on rslearn datasets with many windows.

### DIFF
--- a/docs/examples/WindowsFromGeojson.md
+++ b/docs/examples/WindowsFromGeojson.md
@@ -61,7 +61,7 @@ with open(os.path.join(ds_path, "source_data/subset/latest.geojson"), "w") as f:
 Now we can use the `dataset add_windows` command to create the windows.
 
 ```
-rslearn dataset add_windows --group default --fname $DATASET_PATH/source_data/subset/latest.geojson --grid_size 256 --utm --resolution 10 --start 2024-01-01T00:00:00+00:00 --end 2024-07-01T00:00:00+00:00 --root dataset/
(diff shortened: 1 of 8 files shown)


APatrickJ (approved):
This looks like a great improvement.

PR #206, opened 2025-07-11
hardcode ZENODO_FILES_DATA in WorldCerealConfidences 

gabrieltseng:
This prevents the following:

```
requests.exceptions.HTTPError: 429 Client Error: TOO MANY REQUESTS for url: https://zenodo.org/api/deposit/depositions/7875105/files
```

Since any changes to the files changes the `ZENODO_RECORD_ID`, I think this is safe to do.

### DIFF
--- a/rslearn/train/tasks/segmentation.py
+++ b/rslearn/train/tasks/segmentation.py
@@ -46,6 +46,7 @@ class SegmentationTask(BasicTask):
         f1_metric_thresholds: list[list[float]] = [[0.5]],
         metric_kwargs: dict[str, Any] = {},
         miou_metric_kwargs: dict[str, Any] = {},
+        prob_scales: list[float] | None = None,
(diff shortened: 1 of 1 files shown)


favyen2 (changes requested):
In other data sources that require something similar, we cache the data to a file and then read it in the parallel workers so that as long as `--no-use-initial-job` is NOT passed then it will save the file on the initial job and then afterward they workers will just read from the file.

For example, the first call to LocalFiles.list_items writes a summary.json showing the bounds of all of the raster files and then successive calls can read that instead of opening a lot of GeoTIFFs to get the metadata.

I think it is a bit different here since there are already a fixed set of files you are expecting, so I think it is fine to hardcode if you prefer, but I would cut it down to just the minimal amount of data we need (filename and download link) instead of keeping the entire response.

PR #207, opened 2025-07-11
Finetuning on concatenated datasets

ryspark:
Finetuning on multiple datasets in memory at once.

- [X]  Collect test metrics at the end of a model test run
- [X]  Metrics, logging, checkpointing across different dataset heads
- [X]  Training/evaluation across different datasets
- [X]  Alternating batches between datasets
- [X] Fast distributed data loader that works across multiple ranks/processes
- [ ]  Mixed-dataset in a single batch
- [ ]  Multiple tasks within a single dataset

Currently only allows one task per dataset - this is the highest priority next fix, since some datasets (ex: v2_pastis) have ambiguous task names like detect or classify that collide with other datasets.

Occasionally there are NCCL hangs at the end of validation loops, this has been hard to reproduce/debug in isolation so unclear if this is due to multi-dataset training or a larger issue.

Also adds support for [ALPA](https://arxiv.org/html/2503.11335v2), but this isn't recommended as the performance gains during backward are quite small. Unclear why this is, maybe something to do with switching off gradients at such a finegrained parameter level?

### DIFF
--- a/rslearn/main.py
+++ b/rslearn/main.py
@@ -8,7 +8,6 @@ from collections.abc import Callable
 from datetime import datetime, timedelta, timezone
 from typing import Any, TypeVar
 
-import lightning as L
 import tqdm
 from lightning.pytorch.cli import LightningArgumentParser, LightningCLI
 from rasterio.crs import CRS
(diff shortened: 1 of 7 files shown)


favyen2 (approved):
This implementation is great! Having MultiDatasetDataModule create one data loader over a MultiDataset of ModelDatasets makes sense to me. Left a few minor comments.

PR #208, opened 2025-07-14
Enable classification task to output raster

yawenzzzz:
This is a tentative approach to enable the Kenya/Nandi crop type inference, which let the classification task to produce a raster. This is a feasible one, but as discussed with Favyen, I think it's better to formulate this as a segmentation task so that output raster is more natural. Also, for segmentation task, it will be nice to input a larger window, and output the corresponding geotiff directly, but this will require a new finetune experiment.

### DIFF
--- a/rslearn/models/multitask.py
+++ b/rslearn/models/multitask.py
@@ -3,6 +3,7 @@
 from typing import Any
 
 import torch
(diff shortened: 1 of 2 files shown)


yawenzzzz:
Resolved in #223

PR #209, opened 2025-07-15
Implement attentive pooling for finetuning

ryspark:
Adds a simple attention mechanism on top of the Helios encoder (instead of simply averaging over modalities and bandsets, compute a weighted average over those dimensions, where the weights are a learned linear projection of Helios embeddings). This hasn't worked so well (gives comparable or worse results to plain finetuning) but the infrastructure is there for more sophisticated experiments.

This was built on top of the code from the `singletask` branch so should be merged after #207.

### DIFF
--- a/rslearn/models/multitask.py
+++ b/rslearn/models/multitask.py
@@ -5,6 +5,7 @@ from typing import Any
 import torch
 
 from rslearn.models.pooling import AveragePool, BasePool
+from rslearn.train.lightning_module import RestoreConfig
 
 
 def apply_decoder(
@@ -93,11 +94,17 @@ class MultiTaskModel(torch.nn.Module):
 
         if checkpoint_path is not None:
             print(f"INFO: loading full model weights from {checkpoint_path}")
-            state_dict = torch.load(checkpoint_path)["state_dict"]
-            self.load_state_dict(
-                {k.replace("model.", "", 1): v for k, v in state_dict.items()},
-                strict=False,
+            restore_config = RestoreConfig(
+                restore_path=checkpoint_path,
+                selector=["state_dict"],
+                remap_prefixes=[("model.", "")],
             )
+            state_dict = restore_config.get_state_dict()
+            result = self.load_state_dict(state_dict, strict=False)
+            if result.missing_keys:
+                print(f"WARNING: missing keys: {result.missing_keys}")
+            if result.unexpected_keys:
+                print(f"WARNING: unexpected keys: {result.unexpected_keys}")
 
         for name, decoder in decoders.items():
             if not isinstance(decoder[0], BasePool):
--- a/rslearn/models/multitask.py
+++ b/rslearn/models/multitask.py
@@ -33,7 +33,14 @@ def apply_decoder(
     # First, apply all but the last module in the decoder to the features
     cur = features
     for module in decoder[:-1]:
-        cur = module(cur, inputs)
+        if isinstance(module, BasePool):
+            cur, weights = module(cur, inputs, return_weights=True)
+        else:
+            cur = module(cur, inputs)
+    print("=" * 100)
+    print("ATTN MEAN:", weights.mean(0))
+    print("ATTN 0:", weights[0, :])
+    print("=" * 100)
 
     if targets is None:
         cur_targets = None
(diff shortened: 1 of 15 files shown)


ryspark:
This is coupled to a corresponding [PR](https://github.com/allenai/rslearn_projects/pull/168) in `rslearn_projects` that changes the output signature of the Helios encoder, not sure if this is a breaking change.

rslearn/models/pooling.py
```diff
@@ -0,0 +1,97 @@
...
+    """Simple average pool over the modalities, so we get one BCHW feature map."""
+
+    def __init__(self) -> None:
+        """Initialize a new ModalityPool."""
+        super().__init__()
+
+    def forward(
+        self,
+        features: list[torch.Tensor],
+        inputs: list[dict[str, Any]],
```
Hgherzog, line 26: Suggestion: Remove unused parameter inputs

rslearn/models/pooling.py
```diff
@@ -0,0 +1,97 @@
...
+        return_weights: bool = False,
+    ) -> tuple[list[torch.Tensor], torch.Tensor] | list[torch.Tensor]:
+        """Pool over the modalities.
+
+        Args:
+            features: 1-list of M x B x C x H x W x T tensor
```
Hgherzog, line 32: Nit: I think D for feature dimension is clearer than C for channels

rslearn/models/multitask.py
```diff
@@ -15,7 +16,10 @@ def apply_decoder(
     name: str,
     outputs: list[dict[str, Any]],
     losses: dict[str, torch.Tensor],
-) -> tuple[list[dict[str, Any]], dict[str, torch.Tensor]]:
+    weights_dict: dict[str, torch.Tensor] | None = None,
+) -> tuple[
```
Hgherzog, line 20: Suggestion: Have this function return none instead so it is clear that the intention is to modify inputs in place rather then to return new objects.

rslearn/train/lightning_module.py
```diff
@@ -208,7 +231,12 @@ def training_step(
         """
         inputs, targets, _ = batch
         batch_size = len(inputs)
-        _, loss_dict = self(inputs, targets)
+        if self.model.__class__.__name__ == "MultiTaskModel":
```
Hgherzog, line 234: Suggestion: Move the RestoreConfig to a utils file or a new .py file. As you know this is pretty brittle as a change in that class name will cause this to break.

rslearn/train/lightning_module.py
```diff
@@ -240,7 +268,14 @@ def validation_step(
         """
         inputs, targets, _ = batch
         batch_size = len(inputs)
-        outputs, loss_dict = self(inputs, targets)
+        if self.model.__class__.__name__ == "MultiTaskModel":
```
Hgherzog, line 271: Same as above

rslearn/models/pooling.py
Hgherzog, line 1: Add some unit tests for these

favyen2 (changes requested):
The functionality to log the cross-attention weights feels too specific to incorporate into the general-purpose RslearnLightningModule and MultiTaskModel. The AveragePool and AttentivePool implementations are good but as-is they operate over MBCHWT tensors making them incompatible with any models except the updated Helios model, which isn't in rslearn.

I think it makes sense to move most of the changes to rslearn_projects:
* Given that the AveragePool and AttentivePool are Helios-specific (at least for now), they would probably fit better in `rslp.helios`. It may make sense for the Helios model to accept an argument pooling_method that defaults to AveragePool, instead of connecting the pooling via a separate model component; this way, the Helios model still ends up returning a BCHW tensor.
* I think the RslearnLightningModule logging customizations would also be better as a subclass WeightLoggingLM of RslearnLightningModule, possibly also in `rslp.helios`. It would just need to override the training_step/validation_step, there will be some duplication there but I don't think it is avoidable. I'm not sure what the best solution for returning the cross-attention weights from the pooling layer is though. Maybe the return types from the encoder/decoder layers should be made more flexible (e.g. dicts) so that they can return additional information which could optionally be used by a custom LightningModule. That would be a more significant overhaul though (requires changes to all of the model components) -- in the meantime, it may be easier to also subclass WeightLoggingMultiTaskModel of MultiTaskModel with the customizations needed to return the cross-attention weights.
* I don't think AveragePool should return weights. There is a lot of customization needed to log weights and I think it makes sense for the user to only enable that when using AttentivePool with Helios, and avoid having an awkward special case in AveragePool to return a placeholder weights of ones. It might make sense for the `return_weights` to be an argument to the constructor for AttentivePool (instead of argument to the forward pass), and the `WeightLoggingMultiTaskModel` can simply assume that the encoder returns weights.

favyen2:
Might be worth discussing how to go about incorporating this at our Helios meeting on Monday morning.

PR #210, opened 2025-07-16
Remove redundant checkpoint restoring code from MultiTaskModel

favyen2:
Checkpoint restoring can be configured via the `restore_config: RestoreConfig` argument to `RslearnLightningModule`. See for example https://github.com/allenai/rslearn_projects/blob/18e1ea18c62f60ed5bad231a61a5c8ac92d4c126/data/satlas/solar_farm/config.yaml#L31

So I think the `checkpoint_path` argument to `MultiTaskModel` is not needed. This PR removes that argument.

Let me know if there are actually some cases that the existing code in RslearnLightningModule cannot handle. In that case I still think it would be best to expand the options for RestoreConfig rather than having checkpoint restore code in each model implementation.

### DIFF
--- a/rslearn/models/multitask.py
+++ b/rslearn/models/multitask.py
@@ -4,8 +4,6 @@ from typing import Any
 
 import torch
 
-from rslearn.train.lightning_module import RestoreConfig
-
 
 def apply_decoder(
     features: list[torch.Tensor],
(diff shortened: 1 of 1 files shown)


favyen2:
Thanks for reviewing, no problem I wasn't clear about the RestoreConfig and we need to add documentation for the various options in the model configuration file eventually.

PR #211, opened 2025-07-16
Add cosine annealing LR schedule

ryspark:
Supporting [cosine annealing schedule](https://docs.pytorch.org/docs/stable/generated/torch.optim.lr_scheduler.CosineAnnealingLR.html#torch.optim.lr_scheduler.CosineAnnealingLR) from torch. Am rerunning some finetuning runs to get a more consistent benchmark, since LR doesn't depend on model performance with this scheduler.

At some point should probably organize the learning-rate-related arguments in `RslearnLightningModule`, I left the structure as-is as to not break configs in `rslearn_projects`.


ryspark:
Closing in favor of #201, once merged new schedulers will be added via config.

PR #214, opened 2025-07-21
Better time series support in the model configuration

### LINKED ISSUE #174: Time series data
Design for handling time series data including both single-pixel and patches.

Current issues include:
- Lots of overhead for single-pixel data especially when each timestep is a separate 1x1 GeoTIFF.
- Lots of redundancy in the model configuration file when loading all groups in a layer.

favyen2:
Improve time series support in the model configuration file.

Previously, if we have a layer that is an image time series (contains multiple sub-layers over time), then in the model config we would need to have one DataInput per timestep. This PR adds two options in DataInput, load_all_item_groups and load_all_layers.

- load_all_item_groups: this adds all item groups under the specified layer as options of layers to load.
- load_all_layers: this reads all of the layer options concatenated on channel axis instead of randomly picking one option to read.

So the two options can be combined to read image time series. They are still treated as tensors with time concatenated over channel axis (T*CxHxW).

Resolves #174. Depends on #200.

### DIFF
--- a/rslearn/data_sources/aws_sentinel1.py
+++ b/rslearn/data_sources/aws_sentinel1.py
@@ -1,0 +1,142 @@
+"""Data source for Sentinel-1 on AWS."""
+
+import os
+import tempfile
+from typing import Any
+
+import boto3
+from upath import UPath
+
+from rslearn.config import RasterLayerConfig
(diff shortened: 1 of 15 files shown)


yawenzzzz (approved):
LGTM!

PR #219, opened 2025-07-24
Gradient conflict algorithm + dataloader updates

ryspark:
Add an algorithm to keep gradients pointing in the same direction across different tasks for multi-task learning, as well as add a new dataloader sampling mode.


rslearn/dataset/window.py
ryspark, line 1: This is probably not the best way to go about doing this, seems messy. @favyen2 is there a better place to add the tag and access bbox metadata?

favyen2: It would be best to populate the metadata from a separate script. e.g. https://github.com/allenai/rslearn_projects/blob/master/one_off_projects/convert_satlas_webmercator_to_rslearn/assign_split.py

PR #221, opened 2025-07-28
Label merging + MoE

ryspark:
Allow finetuning a single task-type head for multiple datasets, e.g. one unified decoder head for vessel detection and marine infrastructure detection.
* Supports label merging, i.e. if task 1 has N outputs and task 2 has M outputs, then they are merged and offset appropriately, with the labels being concatenated into a N+M dimensional target
* Models are saved in this label-merged format, see `rslearn_projects` for scripts to unmerge saved models
* Supports task embedding layers to distinguish between different tasks, including a simple per-task learnable embedding lookup and a multi-headed cross-attention layer along the spatial dimensions (Q = per-task learnable embedding + fixed spatial positional encodings, K/V = encoder output – similar to the attention evals in pretraining `helios`)
* **NOTE:** Though sometimes detect/segment tasks have redundant `unknown` or `background` classes, those are kept separate when merging labels between tasks, since an `unknown` from one task might correspond to a known label from another task. Hopefully, this won't make much of a difference due to the task embeddings.

Other new changes (sorry for the long PR):
* Adds support for `DecoderTrunk`, shared weights between the encoder and task-specific decoder heads. This module can inject arbitrary auxiliary losses during training.
* Adds [soft Mixture-of-Experts](https://arxiv.org/abs/2308.00951) as well as infrastructure for extending to other types of decoder trunks.
* For validation with `load_all_patches=False`, the random patch selection per instance is now consistent across different runs (`rng` set based on epoch number).
* **BREAKING:** Changes the signature of `SingleTaskModel` and `MultiTaskModel` to output a dictionary with keys `outputs`, `loss_dict`, and any other keys, instead of a tuple of `(outputs, loss_dict)`. This is useful for passing arbitrary information from the decoders to `RslearnLightningModule`. I also added hooks `on_train_forward`, `on_val_forward`, and `on_test_forward` in `RslearnLightningModule` (by default no-ops) that allow users to access and manipulate this output dictionary (e.g. for custom logging with model weights passed from the decoders, etc).

### DIFF
--- a/rslearn/models/detr/__init__.py
+++ b/rslearn/models/detr/__init__.py
@@ -1,0 +1,5 @@
+"""DETR object detection model code."""
+
+from .detr import Detr
+
+__all__ = ["Detr"]
--- a/rslearn/models/detr/box_ops.py
+++ b/rslearn/models/detr/box_ops.py
@@ -1,0 +1,103 @@
+"""Utilities for bounding box manipulation and GIoU.
+
+This is copied from https://github.com/facebookresearch/detr/.
+The original code is:
+Copyright (c) Facebook, Inc. and its affiliates. All Rights Reserved
+"""
+
+import torch
+from torchvision.ops.boxes import box_area
+
+
+def box_cxcywh_to_xyxy(x: torch.Tensor) -> torch.Tensor:
+    """Convert boxes from cxcywh format to xyxy format."""
+    x_c, y_c, w, h = x.unbind(-1)
+    b = [(x_c - 0.5 * w), (y_c - 0.5 * h), (x_c + 0.5 * w), (y_c + 0.5 * h)]
+    return torch.stack(b, dim=-1)
+
+
+def box_xyxy_to_cxcywh(x: torch.Tensor) -> torch.Tensor:
+    """Convert boxes from xyxy format to cxcywh format."""
+    x0, y0, x1, y1 = x.unbind(-1)
+    b = [(x0 + x1) / 2, (y0 + y1) / 2, (x1 - x0), (y1 - y0)]
+    return torch.stack(b, dim=-1)
+
+
+# modified from torchvision to also return the union
+def box_iou(
+    boxes1: torch.Tensor, boxes2: torch.Tensor
+) -> tuple[torch.Tensor, torch.Tensor]:
+    """Compute the intersection-over-union score between the two lists of boxes.
+
+    The boxes should be in xyxy format.
+
+    Args:
+        boxes1: the first list of boxes (Nx4).
+        boxes2: the second list of boxes (Mx4).
+
+    Returns:
+        the intersection-over-union score.
+    """
+    area1 = box_area(boxes1)
+    area2 = box_area(boxes2)
+
+    lt = torch.max(boxes1[:, None, :2], boxes2[:, :2])  # [N,M,2]
+    rb = torch.min(boxes1[:, None, 2:], boxes2[:, 2:])  # [N,M,2]
+
+    wh = (rb - lt).clamp(min=0)  # [N,M,2]
+    inter = wh[:, :, 0] * wh[:, :, 1]  # [N,M]
+
+    union = area1[:, None] + area2 - inter
+
+    iou = inter / union
+    return iou, union
+
+
+def generalized_box_iou(boxes1: torch.Tensor, boxes2: torch.Tensor) -> torch.Tensor:
+    """Generalized IoU from https://giou.stanford.edu/.
+
+    The boxes should be in [x0, y0, x1, y1] format
+
+    Returns a [N, M] pairwise matrix, where N = len(boxes1)
+    and M = len(boxes2)
+    """
+    # degenerate boxes gives inf / nan results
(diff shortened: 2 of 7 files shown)


rslearn/models/multitask.py
```diff
@@ -32,34 +35,50 @@ def __init__(
...
             lazy_decode: if True, only decode the outputs specified in the batch.
             loss_weights: weights for each task's loss (default: None = equal weights)
+            task_embedding: if provided, use this task embedding module to add task-specific
+                embeddings to the features (recommended if merging decoder heads)
+            decoder_to_target: mapping from decoder id to list of task names
+                (specify if merging heads, otherwise leave as None)
         """
         super().__init__()
         self.lazy_decode = lazy_decode
+        self.decoder_to_target: dict[str, list[str]] = decoder_to_target  # type: ignore
```
favyen2, line 45: `self.decoder_to_target` could be None here since `decoder_to_target` can be None. I don't think the `type: ignore` should be used here.

rslearn/train/tasks/multi_task.py
```diff
@@ -143,24 +181,84 @@ def __init__(self, task_name: str, metric: Metric):
...
+        Since we use the same pred/target tensors for all metrics in the collection,
+        we need to clone them before modifying them.
+
+        Assume first dimension is the number of outputs.
+
+        Args:
+            pred: the prediction
+            target: the target
+        """
+        offset = self.task_label_offsets[self.task_name]["offset"]
```
favyen2, line 214: What if self.task_label_offsets[self.task_name] is not set? Seems like it should default to 0 or avoid calling separate_task_labels.

favyen2 (commented):
If you have one task with 3 classes and one with 4 classes, I can see how you can use the task embedding to change what gets predicted, but if you want to output 7 probabilities and use the task_label_offsets too then how would you do that? I guess I couldn't figure out how you might increase the number of classes for that decoder since it seems to come from one of the tasks.

rslearn/train/tasks/multi_task.py
ryspark: Will update this, though `task_label_offsets` is a key generated by a script in `rslp` so adheres to a pretty rigid format in practice

ryspark:
> Also, would it make sense (and is it possible) to handle the offsets also within the torch.nn.Module instead of in the Task?

Yeah, this would probably make sense, the main requirements are that inputs/targets are properly merged/unmerged prior to model forward pass and prior to metric computation. I could move this to `MultiTaskModel` or a copy, I didn't originally because I wanted to keep all data operations in `tasks/`. But, if you think this is cleaner, I can refactor it.

> If you have one task with 3 classes and one with 4 classes, I can see how you can use the task embedding to change what gets predicted, but if you want to output 7 probabilities and use the task_label_offsets too then how would you do that? 

What do you mean by outputting 7 probabilities while using `task_label_offsets`– like if we wanted 3 + 4 + 7 total outputs across 3 tasks? This is also probably confusing without the context from `rslp`, but configs using `task_label_offsets` are never really expected to be handwritten, it's just a way to pass information from the individual configs to a merged decoder one. It's a little clunky, but merging the configs programmatically seemed much better than trying to re-initialize decoders on the go.

ryspark:
Updated since the last review:
1) Moved all label merging functionality to `MultiTaskMergedModel` (away from the `MultiTask`), subclassed from `MultiTaskModel`.
2) Added MoE implementation (mostly copied from `https://github.com/lucidrains/soft-moe-pytorch`, see `rslearn/models/moe` directory).
3) Improved `DecoderTrunk` support for post-encoder MoE, task embeddings, etc. (this might make more sense in the encoder at some point, but for now I left it in the decoder so it's not tied to any specific backbone).
4) **BREAKING**: changed the signature of SingleTaskModel and MultiTaskModel to output a dictionary with keys outputs, loss_dict, and any other keys, instead of a tuple of (outputs, loss_dict).

rslearn/train/lightning_module.py
```diff
@@ -345,7 +356,7 @@ def predict_step(
             Output predicted probabilities.
         """
         inputs, _, _ = batch
-        outputs, _ = self(inputs)
+        outputs, *_ = self(inputs)
```
favyen2, line 359: I think it outputs dict now.

rslearn/models/multitask.py
```diff
@@ -85,44 +150,242 @@ def apply_decoder(
...
+            trunk_out = self.trunk(features, inputs)
+            outs = self.apply_decoders(trunk_out.pop("outputs"), inputs, targets)
+            self.trunk.apply_auxiliary_losses(trunk_out, outs)
+            return outs | trunk_out
+        else:
+            return self.apply_decoders(features, inputs, targets)
+
+
+class MultiTaskMergedModel(MultiTaskModel):
+    """MultiTaskMerged model.
```
favyen2, line 238: Instead of repeating the docstring it should focus on the difference from MultiTaskModel (merging part) with some detail about what it means to merge and why the user might want to do that.

favyen2 (approved):

PR #231, opened 2025-08-10
Adding Temporal Compositing Options

rohansaw:
**Summary**
This PR adds functionality to create mean and median composites.
 
**Changes**
- Query Config spatial mode now includes `COMPOSITE` -> creates single group for evrey window containing all items in time range.
- RasterLayer Config now includes `compositing_method` option, allowing to specify `FIRST_VALID | MEAN | MEDIAN`.

**Discussion Points**
- Mean compositing: Currently quite memory inefficient: Loads all items windows of the time-period into memory. We might want to think about doing more of a running mean / running-sum approach to reduce memory requirements. Did a short benchmark and running-mean was on average 45% slower however reduced memory significantly depending on the number of items.
- Max_matches: Compositing is currently not using `max_matches` as I believe only a single composite should be created per time period. However, max_matches in this case could be used to limit the number of items in the composite. This might be helpful e.g for having items sorted by cloud cover and then building the composite only with e.g top-k matches. However, would make the implementation a bit more complex as we would have to identify the top-k matches per pixel.
- Instead of the `COMPOSITE` spatial mode that I added, this might be kept as `MOSAIC`. However from my understanding for a composite the `MOSAIC` mode would be required setting `max_matches` > 1, which is interpreted as having multiple final mosaics instead of multiple items included in the mosaic and would populate multiple groups.

### DIFF
--- a/docs/DatasetConfig.md
+++ b/docs/DatasetConfig.md
@@ -143,7 +143,10 @@ Raster layers have additional configuration:
   // layers with a data source. It is used when there is a difference in CRS or
   // resolution between the item from the data source and the window's target.
   // It is one of "nearest", "bilinear" (default), "cubic", "cubic_spline".
-  "resampling_method": "bilinear"
+  "resampling_method": "bilinear",
+  // Method how to select pixel values when multiple items in a group cover the same
+  // pixel. One of "FIRST_VALID", "MEAN", "MEDIAN". Defaults to "FIRST_VALID".
+  "compositing_method": "FIRST_VALID"
 }
 ```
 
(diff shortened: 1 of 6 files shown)


favyen2 (commented):
Thanks for this PR -- this looks great! Just left one comment about the calls to build_X_composite.

The composite/mosaic could be combined into one option (like mosaic_compositing) that controls the desired number of overlaps within one mosaic, where the current mosaic corresponds to mosaic_compositing=1, but I think the current implementation is good too. I wouldn't use max_matches for this since I think that is assumed in a few places to be the maximum number of item groups.

rohansaw:
> Thanks for this PR -- this looks great! Just left one comment about the calls to build_X_composite.
> 
> The composite/mosaic could be combined into one option (like mosaic_compositing) that controls the desired number of overlaps within one mosaic, where the current mosaic corresponds to mosaic_compositing=1, but I think the current implementation is good too. I wouldn't use max_matches for this since I think that is assumed in a few places to be the maximum number of item groups.

Yup, agreed - I also would try to not mix the usage of max_matches like that. For now I'd just leave it as it is for the sake of simplicity and once we get to a point where support for a specific number of overlaps in the composite is required, that could be added.

favyen2 (approved):

PR #244, opened 2025-08-21
Remove __len__ for AllPatchesDataset so prediction runs to completion

favyen2:
`AllPatchesDataset.__len__` was added back but it causes only a subset of patches to be processed.

Previously we tried to implement `__len__` correctly but the length changes based on the number of worker processes since with more workers we end up needing more padding. But pytorch lightning cuts off the prediction based on the length returned when it hasn't yet spawned the worker processes. So it is best to omit it.

Example:
- Dataset contains one window with 16 patches.
- In main process, for batch size 4, length call returns 4 batches.
- If there are four data loader worker processes, Lightning computes 1 batch per process.
- But since we split windows across worker processes, we still need 4 batches since one worker gets the one window and the others need to be padded to the 4 batches.
- So then Lightning cuts off the prediction early because it limits to its estimated 1 batch per process. The output for the window doesn't get generated since not all patches were seen.

I added a test to prevent the bug in the future.

### DIFF
--- a/tests/integration/train/test_prediction_writer.py
+++ b/tests/integration/train/test_prediction_writer.py
@@ -3,12 +3,14 @@
 import lightning.pytorch as pl
 
 from rslearn.dataset.dataset import Dataset
+from rslearn.models.conv import Conv
+from rslearn.models.module_wrapper import EncoderModuleWrapper
 from rslearn.models.multitask import MultiTaskModel
 from rslearn.models.pooling_decoder import PoolingDecoder
 from rslearn.models.singletask import SingleTaskModel
 from rslearn.models.swin import Swin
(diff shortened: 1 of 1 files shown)


favyen2:
I tried a few ways to compute the length but in the way it seems impossible unless we know the number of workers a priori (which we could estimate if we pass number of workers from DataModule to AllPatchesDataset but this feels error-prone).

* https://github.com/allenai/rslearn/pull/223/commits/0f40f6df7c3199b5a10f9ae3f9be6988aa060bff
* https://github.com/allenai/rslearn/pull/223/commits/877d5cd455eba0ad4b8e750fa858257f02801ed3

PR #247, opened 2025-08-25
Add EuroCrops data source and refresh the data source documentation

favyen2:
Add EuroCrops data source and refresh the data source documentation.


rslearn/data_sources/eurocrops.py
```diff
@@ -0,0 +1,246 @@
...
+        return EuroCrops()
+
+    def _get_all_items(self) -> list[EuroCropsItem]:
+        """Get a list of all available items in the data source."""
+        items: list[EuroCropsItem] = []
+        for year, fnames in self.FILENAMES_BY_YEAR.items():
+            items.append(
+                EuroCropsItem(
+                    str(year),
+                    get_global_geometry(
```
yawenzzzz, line 125: Just curious, given that this dataset is for EU, why using the global geometry?

favyen2: I think the EU polygon would have issues with re-projecting to some projections due to its large geographic extent.

Also since the whole data source is EuroCrops there isn't a risk of matching one of these items to a window instead of another better one outside the EU, the worst that would happen is creating empty GeoJSONs for those windows.

PR #253, opened 2025-08-28
Support configurable output selector and custom output path in RslearnWriter

### LINKED ISSUE #251: Support configurable output selector and custom output path in RslearnWriter
Currently, `RslearnWriter` is fixed to use `selector=["class"]`, which causes issues when using checkpoints trained with different task names (e.g., segment). This requires manual remapping of state dict keys, such as:

```
import torch

state_dict = torch.load("checkpoint.ckpt")
state_dict["state_dict"] = {
    k.replace("model.decoders.segment.", "model.decoders.class."): v
    for k, v in state_dict["state_dict"].items()
}
torch.save(state_dict, "checkpoint.ckpt")
```

It would be more flexible if the selector could be configured at runtime to match the task name, or standardized to a generic identifier like "`task`".

Additionally, there’s a related need to:
	•	Allow writing predictions to an arbitrary path rather than being tied strictly to the dataset path.
	•	Pass in a layer config so that dataset config under that path is not required.

This plays into recent work (e.g., Sebastian’s PR exploring a separate writer), but instead of creating entirely new writers, it seems feasible to incorporate these features directly into `RslearnWriter `without breaking existing functionality.

⸻

Proposed Changes
	•	Add two new optional arguments to `RslearnWriter`:
	1.	`selector` (list[str]): configurable keys for selecting task output, defaulting to `["class"]` for backward compatibility.
	2.	`output_path` (str | Path): optional path for writing predictions, instead of always resolving under dataset path.
	•	Ensure existing functionality is preserved when these arguments are not provided.

⸻

Rationale
	•	Removes the need for checkpoint hacks when task names differ (`segment` vs `class`).
	•	Provides greater flexibility for downstream use cases without requiring custom writer subclasses.
	•	Aligns with the goal of making `RslearnWriter` more general-purpose rather than fragmented by task-specific writers.

joshhvulcan:
Closes #251

### DIFF
--- a/rslearn/train/prediction_writer.py
+++ b/rslearn/train/prediction_writer.py
@@ -2,6 +2,7 @@
 
 from collections.abc import Sequence
 from dataclasses import dataclass
+from pathlib import Path
 from typing import Any
 
 import numpy as np
@@ -125,13 +126,17 @@ class RslearnWriter(BasePredictionWriter):
     for each window being processed.
     """
 
+    layer_config: RasterLayerConfig | VectorLayerConfig
+
     def __init__(
         self,
         path: str,
         output_layer: str,
-        path_options: dict[str, Any] = {},
-        selector: list[str] = [],
+        path_options: dict[str, Any] | None = None,
+        selector: list[str] | None = None,
         merger: PatchPredictionMerger | None = None,
+        output_path: str | Path | None = None,
+        layer_config: RasterLayerConfig | VectorLayerConfig | None = None,
     ):
         """Create a new RslearnWriter.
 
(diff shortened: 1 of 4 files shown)


rslearn/train/prediction_writer.py
```diff
@@ -125,13 +126,17 @@ class RslearnWriter(BasePredictionWriter):
     for each window being processed.
     """
 
+    layer_config: RasterLayerConfig | VectorLayerConfig
```
favyen2, line 129: It seems like this type hint should be instance-level instead of here (`self.layer_config: RasterLayerConfig | VectorLayerConfig` in `__init__`).

rslearn/train/prediction_writer.py
```diff
@@ -231,9 +263,13 @@ def process_output_batch(
             for k in self.selector:
                 output = output[k]
 
+            # # Use custom output_path if provided, otherwise use dataset path
```
favyen2, line 266: Extra #

tests/unit/train/test_prediction_writer.py
```diff
@@ -188,3 +260,553 @@ def test_write_raster(tmp_path: pathlib.Path) -> None:
...
+    assert writer_custom.selector == ["segment"], "Should use custom selector"
+
+
+def test_selector_with_dictionary_output(tmp_path: pathlib.Path) -> None:
+    """Test RslearnWriter selector functionality with dictionary outputs.
+
+    Tests that selector=['segment'] correctly extracts the 'segment' key from
+    task outputs and that the codepath in process_output_batch() is covered.
+    """
+    from rslearn.config import BandSetConfig, RasterLayerConfig
```
favyen2, line 502: Is there reason not to put this import at the top of file? Same for the other imports like Mock.

favyen2 (approved):

tests/unit/train/test_prediction_writer.py
```diff
@@ -188,3 +260,553 @@ def test_write_raster(tmp_path: pathlib.Path) -> None:
...
+        "group": window_group,
+        "bounds": (0, 0, 5, 5),
+        "window_bounds": (0, 0, 5, 5),
+        "projection": Projection(WGS84_PROJECTION.crs, 0.2, 0.2),
+        "time_range": None,
+        "patch_idx": 0,
+        "num_patches": 1,
+    }
+
+    # Create model output - 3 classes, 5x5 spatial dimensions
```
favyen2, line 552: This information is not used in the test. I think it would be simpler to just set to torch.zeros instead of all the probability assignments.

tests/unit/train/test_prediction_writer.py
```diff
@@ -188,3 +260,553 @@ def test_write_raster(tmp_path: pathlib.Path) -> None:
...
+        time_range=None,
+    )
+    expected_fname = (
+        window.get_raster_dir(output_layer_name, output_bands, 0) / "geotiff.tif"
+    )
+    assert expected_fname.exists(), "Output should be written with custom layer config"
+    assert window.is_layer_completed(output_layer_name)
+
+
+def test_selector_configuration(tmp_path: pathlib.Path) -> None:
```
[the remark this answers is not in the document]
joshhvulcan: Removing!

PR #254, opened 2025-08-28
Add map-style all patches dataset

ryspark:
Renames the current implementation of `AllPatchesDataset` to `IterableAllPatchesDataset` and adds a new map-style implementation, `InMemoryAllPatchesDataset`, with `__getitem__` and `__len__`.
* Eliminates the need for custom distributed logic, since it can be handled by `DistributedSampler` in the data module
* Necessary for running evals locally on multi-task models with `model validate`
* After some very basic testing it's pretty much the same speed as the iterable version
* By default we use the iterable version to avoid breaking large validation dataset runs, but users can switch via the  `use_in_memory_all_patches_dataset` argument in `RslearnDataModule`


rslearn/train/dataset.py
```diff
@@ -1066,6 +1066,194 @@ def get_dataset_examples(self) -> list[Window]:
...
+
+        Returns:
+            a tuple of (raw_inputs, passthrough_inputs, metadata).
+        """
+        if index in self.window_cache:
+            return self.window_cache[index]
+
+        # Pad tensors by patch size to protect slicing near right/bottom edges.
+        return_inputs = self.dataset.get_raw_inputs(index)
+        for d in return_inputs[:-1]:
```
favyen2, line 1194: The padding and cropping code should be shared as well.

rslearn/train/dataset.py
```diff
@@ -1066,6 +1066,194 @@ def get_dataset_examples(self) -> list[Window]:
         return self.dataset.get_dataset_examples()
 
 
+class AllPatchesDataset(torch.utils.data.Dataset):
```
favyen2, line 1095: It would be good to document that this results in the dataset being stored in memory. I would name it `InMemoryAllPatchesDataset`. I don't think it should be the default option, it can break some existing configs where we have large validation set. For a non-iterable dataset for all patches that is the default option, as discussed I think it should not store in memory and instead perform the random access read using `read_data_input` with the patch bounds.

ryspark: Ok - then in the case where `load_all_patches=True` we want to default to the iterable version?

favyen2: If it is in-memory, why not simply wrap the existing iterable AllPatchesDataset and have a very small class that does `self.examples = list(all_patches_dataset)`?

favyen2: The random access is essentially what we had before the iterable AllPatchesDataset addition, but it wasn't working for very small windows like 4x4 or 8x8 (too slow data loading). It was otherwise fine though, but it wasn't cleanly implemented, so it would be cleaner if it can be a separate class (previously we were overloading the type of `self.dataset_examples` to either be `list[Window]` for regular mode and `list[tuple[Window, PixelBounds, int]]` for all patches mode.

rslearn/train/lightning_module.py
```diff
@@ -209,13 +209,14 @@ def on_train_epoch_start(self) -> None:
             # Fail silently for single-dataset case, which is okay
             pass
 
-    def on_test_epoch_end(self) -> None:
+    def on_validation_epoch_end(self) -> None:
```
favyen2, line 212: This will save the metrics on validation epoch instead of test epoch. If this is desired then it should be clearly documented in the constructor (for the `metrics_file` argument). The docstring here would also need update. I thought saving on test made more sense.

PR #262, opened 2025-09-02
Remove unnecessary test fixtures

### LINKED ISSUE #260: test_edc_preview depends on GCS access
Attempting to run `test_edc_preview` fails with GCS credentials error due to 
```
@pytest.fixture(scope="session", autouse=True)
    def test_bucket() -> str:
        os.environ.setdefault("TEST_BUCKET", "test-bucket-rslearn")
        test_bucket = os.environ["TEST_BUCKET"]
        print(f"test_bucket: {test_bucket}")
>       storage_client = storage.Client()
```
I don't have a GCS account so cannot run the test, creating a blocker to contributing to rlsearn. I wonder if a local solution could be used rather than GCS, or something like [minio](https://www.min.io/)?
Thanks

favyen2:
The `test_bucket` and `test_prefix` fixtures seem to just set some defaults for environment variables. It seems better to let the tests that depend on them fail if the environment variables are not set, rather than set defaults that would only work for people with access to that specific bucket.

Resolves #260.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -62,11 +62,6 @@ jobs:
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
 
-      - name: Set Environment Variables
-        run: |
-          echo "TEST_BUCKET=test-bucket" >> $GITHUB_ENV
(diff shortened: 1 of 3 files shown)


APatrickJ (approved):
Great, nice to see this cleanup

PR #263, opened 2025-09-02
Configure EDS env vars in data source integration tests

APatrickJ:


favyen2 (commented):
We set environment variable from secret and then pass to Docker command from environment variable so the escaping can be handled at both points (there doesn't seem to be a good way to escape the secret directly into the command).

But here the environment variables don't seem like they have been set.

favyen2:
Henry set up the data source tests to run twice a week or something like that because it takes an hour to run through all of them. But you can still verify that changes work by manually running the test from https://github.com/allenai/rslearn/actions/workflows/data-source-integration-test.yml

favyen2 (approved):

APatrickJ:
> Henry set up the data source tests to run twice a week or something like that because it takes an hour to run through all of them. But you can still verify that changes work by manually running the test from https://github.com/allenai/rslearn/actions/workflows/data-source-integration-test.yml (select "Run workflow" and then select this branch).

Okay good idea to test it, manually triggered: https://github.com/allenai/rslearn/actions/runs/17409866365

PR #265, opened 2025-09-02
Miscellaneous bug fixes

favyen2:
- Force return_predictions=False to avoid high memory consumption issues
- Fix bug with AllPatchesDataset/RslearnWriter where empty output may be written
- Add logging for RslearnWriter
- Improve speed of SegmentationTask.process_output by keeping on GPU


yawenzzzz (approved):
LGTM!

PR #266, opened 2025-09-02
Add class id remapper

yawenzzzz:
Add a remapper to map old class ids to new ones. In the model config, we can specify sth like this: `class_id_mapping: {1: 1, 2: 2, 3: 2, 4: 2, 5: 2, 6: 0, 7: 0, 8: 0, 9: 0, 10: 0, 11: 2}`

### DIFF
--- a/rslearn/train/tasks/segmentation.py
+++ b/rslearn/train/tasks/segmentation.py
@@ -127,11 +127,12 @@ class SegmentationTask(BasicTask):
 
         assert raw_inputs["targets"].shape[0] == 1
         labels = raw_inputs["targets"][0, :, :].long()
-        new_labels = labels.clone()
 
         if self.class_id_mapping is not None:
+            new_labels = labels.clone()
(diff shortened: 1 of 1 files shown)


rslearn/train/tasks/segmentation.py
```diff
@@ -112,14 +116,19 @@ def process_inputs(
         # TODO: List[Feature] is currently not supported
         assert raw_inputs["targets"].shape[0] == 1
         labels = raw_inputs["targets"][0, :, :].long()
+        new_labels = labels.clone()
+
+        if self.class_id_mapping is not None:
```
favyen2, line 121: I think this logic would be cleaner if it is isolated to this if statement.

```python
if self.class_id_mapping is not None:
    new_labels = labels.clone()
    for old_id, new_id in self.class_id_mapping.items():
                        new_labels[labels == old_id] = new_id
    labels = new_labels
```

favyen2 (approved):

PR #269, opened 2025-09-03
Fix AllPatchesDataset bug when there are zero windows

favyen2:
Fix AllPatchesDataset bug when there are zero windows.

This addresses failing tests in rslearn_projects.


yawenzzzz (approved):
LGTM!

PR #270, opened 2025-09-04
Allows arbitrary nodata_value in segmentation

cmwilhelm:
Previously, SegmentationTasks allowed labels of `0` to 
be recognized as invalid. They would be masked
during loss and metrics calculation.

This changeset allows an arbitrary value to be
specified as a new `nodata_value` init parameter.


favyen2 (approved):
looks good to me, thanks for adding this and adding the tests!

Issue #273, opened 2025-09-12
LocalFiles fails for global raster


robmarkcole:
I have a static, global coverage GeoTIFF in EPSG:4326 representing soil type (`size=(194713,70879), crs=EPSG:4326, transform=… bounds=(-94542, -39392, 100171, 31487)`). Windows are small UTM tiles (e.g., EPSG:32614). After `materialize` other layers are generated, but not the soil type layer. 

Global raster reprojection: 
- The LocalFiles raster importer builds an item geometry in the image’s native CRS; during matching, the item geometry is reprojected into each window’s UTM CRS to run shapely intersections. For a very large global EPSG:4326 raster, reprojection into a narrow UTM zone can produce invalid/degenerate polygons, causing spatial matching to fail and return 0 groups — even though a preliminary pre-filter reported 1/1 items intersecting.
- Missing “global coverage” handling: The Vector importer treats “too large” geometries as global to avoid problematic reprojections, but the Raster importer did not. As a result, the matching step tried to reproject the full-world raster into UTM instead of treating it as globally covering the window.

Fix: RasterImporter parity with VectorImporter

PR #274, opened 2025-09-15
check raster geometry.is_too_large

robmarkcole:
Resolve https://github.com/allenai/rslearn/issues/273

### DIFF
--- a/rslearn/data_sources/local_files.py
+++ b/rslearn/data_sources/local_files.py
@@ -232,6 +232,11 @@ class RasterImporter(Importer):
                 projection = Projection(crs, x_resolution, y_resolution)
                 geometry = STGeometry(projection, shp, None)
 
+            # If the raster item has very large spatial coverage, projections with
+            # limited validity (e.g., UTM) can fail spatial tests. Treat as global.
+            if geometry.is_too_large():
+                geometry = get_global_geometry(time_range=None)
+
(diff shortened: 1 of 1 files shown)


robmarkcole:
@favyen2 any further clarification required?

favyen2:
I don't think so, but I was trying to reproduce this and so far wasn't able to although I have seen similar errors in the past, and I was also trying to reproduce the rasterio not producing error when response is truncated but so far every way I've tried it does raise an exception if the connection is closed.

So I will test some more but if you want to make a PR with the global geometry as a separate thing I think that would be quicker/easier to merge in. It may be better to add a logger.warning for that global geometry change though, since if the geometry is treated as global, then it will be matched with all windows in the rslearn dataset, and e.g. when using max_matches=1, space_mode=MOSAIC it may use the global raster to materialize data for a window that is outside its actual valid bounds instead of another more appropriate raster. It would work fine with other modes like COMPOSITE though, or if the user just sets max_matches really high and deals with the issue themselves.

robmarkcole:
@favyen2 I will create a separate PR for global geometry change

Regarding the Planetary computer errors, it's a bit suspect if I'm the only one reporting it. Perhaps keep those issues open and see if anyone else hits them. If not perhaps its a quirk of my VPN setup or something not easy to reproduce

Issue #276, opened 2025-09-16
ERA5 monthly ingest fails with MultiAdaptorNoDataError


robmarkcole:
With a window metadata.json
```json
{
    "group": "global_soil_moisture",
    "name": "Anthony_37.2134_-98.0969_2016-07-05",
    "projection": {
        "crs": "EPSG:32614",
        "x_resolution": 10,
        "y_resolution": -10
    },
    "bounds": [
        58009,
        -411896,
        58017,
        -411888
    ],
    "time_range": [
        "2016-07-04T00:00:00+00:00",
        "2016-07-06T00:00:00+00:00"
    ],
    "options": {
        "split": "val",
        "site_name": "Anthony",
        "latitude": 37.2134,
        "longitude": -98.0969
    }
}
```
On ingest:
```
format=2025-09-16 12:26:13,978 loglevel=INFO   logger=rslearn.main __call__() L512  Running ingest for 1 jobs
format=2025-09-16 12:26:13,978 loglevel=DEBUG  logger=rslearn.data_sources.climate_data_store ingest() L290  CDS API request for the whole globe for year=['2016'] month=['07']
format=2025-09-16 12:26:13,978 loglevel=DEBUG  logger=rslearn.data_sources.climate_data_store ingest() L290  CDS API request for the whole globe for year=['2016'] month=['07']
2025-09-16 12:26:14,168 INFO Request ID is 2d496604-870e-45b6-87b0-ab3e92ea0d90
2025-09-16 12:26:14,242 INFO status has been updated to accepted
2025-09-16 12:26:22,800 INFO status has been updated to running
2025-09-16 12:26:27,940 INFO status has been updated to failed
format=2025-09-16 12:26:28,076 loglevel=WARNING logger=rslearn.dataset.manage retry() L45   Retrying after catching error in retry loop: 400 Client Error: Bad Request for url: https://cds.climate.copernicus.eu/api/retrieve/v1/jobs/2d496604-870e-45b6-87b0-ab3e92ea0d90/results
The job has failed
The job failed with: MultiAdaptorNoDataError
```

robmarkcole:
OK new issue when using 
```
        "era5": {
            "type": "raster",
            "band_sets": [
                {
                    "dtype": "float32",
                    "bands": [
                        "total_precipitation"
                    ]
                }
            ],
            "data_source": {
                "name": "rslearn.data_sources.climate_data_store.ERA5LandMonthlyMeans",
                "api_key": "mine"
            }
        }
[...]

PR #277, opened 2025-09-16
update era5 info

robmarkcole:
Addresses https://github.com/allenai/rslearn/issues/276
Note I updated the url since the previously referenced url returns 504 error


APatrickJ (approved):
Thank you for the update!

PR #283, opened 2025-09-19
Fixes environment variable substitution.

cmwilhelm:
The previous implementation was intercepting
argument processing too late. Although the substitution logic worked
in isolation on a `Namespace` object, the upstream processing on the
configuration object performs type validation (i.e. init_args type
matching for class path constructor signatures)
that we violate with late substitution.

This changeset implements substitution as a LightningArgumentParser,
intercepting the raw yaml string before it is parsed. 
This requires RslearnLightningCLI to be instantiated with the new parser class.


yawenzzzz (approved):
LGTM!

PR #285, opened 2025-09-22
Add AnySat

yawenzzzz:
The AnySat model is ready, we may need to remove some of the multimodal tests to speed up the test, it's a little bit slow right now. The author mentioned "The data also needs to be normalized (data -mean) /std as input to our model", and there's no specific norm stats provided.


rslearn/models/anysat.py
```diff
@@ -0,0 +1,238 @@
...
+        Args:
+            spatial_shapes: dict mapping modality to (H, W) of that modality in the batch.
+        """
+        required_ps_m = self.patch_size_meters
+        for m, (H, W) in spatial_shapes.items():
+            res_m = MODALITY_RESOLUTIONS[m]
+            need_m = math.ceil((res_m * max(H, W)) / 32.0)
+            if need_m > required_ps_m:
+                required_ps_m = need_m
+        self.patch_size_meters = self._ceil_to_multiple(required_ps_m, 10)
```
favyen2, line 140: It feels odd that this would change the field rather than just being applied for each forward pass.

favyen2 (approved):

rslearn/models/anysat.py
yawenzzzz: OK, i added the `adjusted_patch_size_meters` so that we only adjust the patch size once, and no change on the original `patch_size_meters`

rslearn/models/anysat.py
```diff
@@ -197,9 +197,8 @@ def forward(self, inputs: list[dict[str, Any]]) -> list[torch.Tensor]:
                     "All modalities must share the same spatial extent (H*res, W*res)."
                 )
 
-        if not self.is_patch_size_adjusted:
+        if self.adjusted_patch_size_meters == 0:
```
favyen2, line 200: what if different batches have different sizes?

unless anysat requires all batches to have same size, or requires same patch_size(meters) across batches, then it seems like you can just compute this per-forward-pass.

rslearn/models/anysat.py
```diff
@@ -113,12 +113,12 @@ def __init__(
         )
         self._embed_dim = 768  # base width, 'dense' returns 2x
         # Assuming all batches have the same spatial shapes, adjust patch size only once
-        self.is_patch_size_adjusted = False
+        self.adjusted_patch_size_meters = 0
```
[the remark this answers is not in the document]
yawenzzzz: yes, we can get adjusted patch size from this function, the reason to store it in field is for the `get_backbone_channels` to get the final `patch_size` as the downscale factor.

rslearn/models/anysat.py
yawenzzzz: I think AnySat does require samples to have fixed shapes, for example, the dates field must apply to all samples so they share the same number of timesteps, and spatially, samples from the same modality always have the same shape (h × w).

favyen2: Is that within a batch or across batches? I was thinking more about across batches, like one batch may have one size/patch_size and next batch a different size/patch_size.

yawenzzzz: Ah, i see what you mean, yeah, it's within batches, across batches, we could have different dates as inputs.

PR #292, opened 2025-09-25
Henryh/add panopticon

Hgherzog:


tests/unit/models/test_panopticon.py
```diff
@@ -0,0 +1,138 @@
...
+            "B06",
+            "B07",
+            "B08",
+            "B8A",
+            "B09",
+            "B11",
+            "B12",
+        ]
+    }
+    panopticon = Panopticon(band_order=band_order).to(DEVICE)
```
Hgherzog, line 34: I should add torch.no_grad here

rslearn/models/panopticon_data/sensors/drone.yaml
```diff
@@ -0,0 +1,32 @@
+instrument: 'unknown'
```
favyen2, line 1: Would be useful to have a README in this folder to say where these YAML files came from. It can also say the purpose, i.e. because Panopticon distinguishes different channels by encoding the associated nanometer wavelengths.

rslearn/models/panopticon.py
```diff
@@ -0,0 +1,167 @@
...
+
+    SENTINEL2 = "sentinel2"
+    LANDSAT8 = "landsat8"
+    SENTINEL1 = "sentinel1"
+
+
+class Panopticon(nn.Module):
+    """Class containing the Panopticon model that can ingest MaskedHeliosSample objects."""
+
+    supported_modalities: list[str] = [
```
favyen2, line 30: Since band_order is required, it seems we shouldn't need supported_modalities and can use self.band_order.keys() instead. This also helps so that user is not limited to just the modalities listed here (although still need PanopticonModalities enum to match with the sensor config).

.pre-commit-config.yaml
```diff
@@ -55,7 +55,8 @@ repos:
           - -s
           # Skip B113 request_without_timeout because it has false positives e.g. when
           # timeout is computed by a function.
-          - B101,B311,B324,B108,B104,B310,B405,B314,B608,B113
+          # skip B614 as it dislikes loading from torch hub
+          - B101,B311,B324,B108,B104,B310,B405,B314,B608,B113,B614
```
favyen2, line 59: I think we should keep B614 since it is useful to avoid mistakenly loading weights without weights_only=True. Suggest to add nosec instead if there is issue with torch.hub.load.

PR #295, opened 2025-09-26
Add Presto

gabrieltseng:


rslearn/models/presto/presto.py
```diff
@@ -0,0 +1,237 @@
...
+        super().__init__()
+
+        if pretrained_path is None:
+            pretrained_path = UPath(tempfile.gettempdir(), "rslearn_cache", "presto")
+        if not (UPath(pretrained_path) / MODEL_FILENAME).exists():
+            _ = hf_hub_download(
+                local_dir=UPath(pretrained_path),
+                repo_id=HF_HUB_ID,
+                filename=MODEL_FILENAME,
+                revision="1b97f885969da4e2d5834ca8c92707c737911464",
```
favyen2, line 70: This may be better as a constant with a comment about what it is for.

gabrieltseng: weirdly, the bandit pre-commit check passes if `revision` is a string like this and fails if its a constant. I have added a comment describing what it is.

favyen2: It may be erroneously thinking it is a password/token. Can add a "REVISION = "..." # nosec" comment after the statement to disable the bandit error.

favyen2 (approved):

rslearn/models/presto/presto.py
```diff
@@ -0,0 +1,257 @@
...
+
+        x = rearrange(x, "b t h w d -> (b h w) t d")
+        mask = rearrange(mask, "b t h w d -> (b h w) t d")
+        dynamic_world = rearrange(dynamic_world, "b t h w -> (b h w) t")
+        months = repeat(months, "b t -> (b h w) t", h=h, w=w)
+        if latlons is not None:
+            latlons = rearrange(latlons, "b c h w -> (b h w) c")
+
+        cur_idx = 0
+        while cur_idx < (b * h * w):
```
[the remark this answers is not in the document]
gabrieltseng: Updated

rslearn/models/presto/presto.py
gabrieltseng: I think its because "revision" can be a branch name too, but bandit specifically wants it to be a commit (and I suppose can't check its a commit if its a constant). Its the same issue which is raised (B615). Would you prefer a "# nosec# over the current approach of commenting what the revision is?

PR #298, opened 2025-09-29
make it possible to use band names with underscore in the name

favyen2:


tests/unit/dataset/test_window.py
```diff
@@ -51,3 +53,14 @@ def test_layer_dir_location(empty_window: Window) -> None:
...
+
+def test_underscore_band_name(empty_window: Window) -> None:
+    """Verify that we can have undescore in the raster band name."""
+    raster_dir = empty_window.get_raster_dir("layer", ["_"])
+    GeotiffRasterFormat().encode_raster(
+        raster_dir, WGS84_PROJECTION, (0, 0, 4, 4), np.zeros((1, 4, 4), dtype=np.uint8)
+    )
+    empty_window.mark_layer_completed("layer")
+    assert empty_window.is_layer_completed("layer")
+    assert len(empty_window.get_layer_dir("layer").iterdir()) == 1
```
APatrickJ, line 66: I think we need to change this to:


```suggestion
    assert len(list(empty_window.get_layer_dir("layer").iterdir())) == 1
```

Or do something similar given `iterdir()` returns a generator

Issue #300, opened 2025-09-30
ENHANCEMENT: wandb log training yaml


robmarkcole:
It would be useful to have the training yaml alongside run metrics and other artefacts. In another repo I implemented this in the cli using a callback to upload the yaml artefact to a run (with mlflow), that could work here.

PR #301, opened 2025-09-30
Wandb logging of train config yaml

robmarkcole:
To address https://github.com/allenai/rslearn/issues/300

Validated the config is logged:
image


favyen2:
Currently we do this in rslearn_projects (https://github.com/allenai/rslearn_projects/blob/cf549e54192954ee606de228ce6ae7cb6ad83a29/rslp/lightning_cli.py#L76), with the original intention being to provide minimal functionality on top of Lightning here in case users use different loggers and such.

But I will look into moving some of that to rslearn. Besides the config saving, specifying how to load checkpoint would also be useful.

PR #305, opened 2025-09-30
Pjreddie/add dinov3

pjreddie:


rslearn/models/dinov3.py
```diff
@@ -0,0 +1,172 @@
...
+    DinoV3Models.LARGE_WEB: "dinov3_vitl16_pretrain_lvd1689m-8aa4cbdd.pth",
+    DinoV3Models.HUGE_PLUS_WEB: "dinov3_vith16plus_pretrain_lvd1689m-7c1da9a5.pth",
+    DinoV3Models.FULL_7B_WEB: "dinov3_vit7b16_pretrain_lvd1689m-a955f4.pth",
+}
+
+
+class DinoV3(torch.nn.Module):
+    """DinoV3 Backbones.
+
+    Must have the pretrained weights downloaded in checkpoint_dir for them to be loaded.
```
favyen2, line 40: Maybe can include link to https://github.com/facebookresearch/dinov3?tab=readme-ov-file#pretrained-models or similar URL somewhere so user can know how to go about obtaining access to the pre-trained weights.

rslearn/models/dinov3.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Only takes RGB as input. Expects normalized data (use the below normalizer).
+
+    Uses patch size 16.
+    """
+
+    modalities: list[str] = ["rgb", "s2", "landsat"]
```
favyen2, line 47: Assuming the model does not natively know about these modalities, it may be better for the model to accept a `image_keys: list[str] = ["image"]` argument that user can customize to specify the name of each modality that they want to apply the model on.

Alternatively, since anyway it seems like all of the different modalities are passed through the same model weights, and all the images have to be three channels, you could just always assume the input is a single image here. Then the user can concatenate the image time series from all the different modalities, and use SimpleTimeSeries to pool the features across images. SimpleTimeSeries wraps a per-image model by applying the model on each image in a time series and then applying temporal pooling across the computed feature maps.

favyen2 (approved):

PR #308, opened 2025-10-01
Fix occasional re-projection error when reading vector features

favyen2:
The error seems to sometimes happen when a feature in a GeoJSON falls outside the area of use of the requested projection. It doesn't always happen and I'm not sure what causes the exact error conditions.

Anyway, the solution employed here is to check in WGS84 and use that to determine if we should even attempt the re-projection to the destination projection. In WGS84 we do need to deal with antimeridian splitting but otherwise should work okay.

### DIFF
--- a/rslearn/models/unet.py
+++ b/rslearn/models/unet.py
@@ -17,10 +17,11 @@ class UNetDecoder(torch.nn.Module):
     def __init__(
         self,
         in_channels: list[tuple[int, int]],
-        out_channels: int,
+        out_channels: int | None,
         conv_layers_per_resolution: int = 1,
         kernel_size: int = 3,
         num_channels: dict[int, int] = {},
+        target_resolution_factor: int = 1,
         original_size_to_interpolate: tuple[int, int] | None = None,
     ) -> None:
         """Initialize a UNetDecoder.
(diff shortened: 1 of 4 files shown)


rslearn/utils/geometry.py
```diff
@@ -464,6 +468,60 @@ def split_at_antimeridian(geometry: STGeometry, epsilon: float = 1e-6) -> STGeom
...
+
+def safely_reproject_and_clip(
+    src_geoms: Sequence[STGeometry], dst_geom: STGeometry
+) -> Sequence[STGeometry | None]:
+    """Re-project src_geoms into the projection of dst_geom.
+
+    The resulting geometries will be clipped to dst_geom. If there is no intersection
+    for an src_geom, then the result will be None. The list of results is returned.
+
+    This function addresses issues with directly re-projection (e.g. using
```
APatrickJ, line 484:
```suggestion
    This function addresses issues with directly re-projecting (e.g. using
```

rslearn/utils/geometry.py
```diff
@@ -464,6 +468,60 @@ def split_at_antimeridian(geometry: STGeometry, epsilon: float = 1e-6) -> STGeom
...
+    src_geom.to_projection(dst_geom.projection), which may fail if the source geometry
```
APatrickJ, line 485: Total nit-pick, there is a missing close paren here from the line above :-)

rslearn/utils/geometry.py
```diff
@@ -464,6 +468,60 @@ def split_at_antimeridian(geometry: STGeometry, epsilon: float = 1e-6) -> STGeom
...
+
+    def check_in_wgs84(src_geom: STGeometry) -> bool:
+        """Return False if there is no intersection."""
+        nonlocal dst_geom_wgs84
+        src_geom_wgs84 = split_at_antimeridian(src_geom.to_projection(WGS84_PROJECTION))
+        if dst_geom_wgs84 is None:
+            dst_geom_wgs84 = split_at_antimeridian(
+                dst_geom.to_projection(WGS84_PROJECTION)
+            )
+        return src_geom_wgs84.intersects(dst_geom_wgs84)
```
APatrickJ, line 506: Optionally, we could simplify this a bit by using the `@cache` decorator:

```python
    @cache
    def get_dst_geom_wgs84() -> STGeometry:
        """Lazily compute and cache dst_geom in WGS84 projection."""
        return split_at_antimeridian(dst_geom.to_projection(WGS84_PROJECTION))
    
    def intersects_in_wgs84(src_geom: STGeometry) -> bool:
        """Return False if there is no intersection."""
        src_geom_wgs84 = split_at_antimeridian(src_geom.to_projection(WGS84_PROJECTION))
        return src_geom_wgs84.intersects(get_dst_geom_wgs84())
```

PR #315, opened 2025-10-03
Pin boto3 and aiobotocore to speed up dependency resolution

joshhvulcan:


pyproject.toml
```diff
@@ -9,7 +9,8 @@ readme = "README.md"
 license = {file = "LICENSE"}
 requires-python = ">=3.11"
 dependencies = [
-    "boto3>=1.39",
+    "aiobotocore==2.24.2",  # Pinning boto and aiobotocore to speed up dependency resolution.
```
favyen2, line 12: Does this need to be pinned here or can it be pinned downstream? If it is needed here, does it work in extra (s3fs is in extra and I only see it used in the aws_X data sources)?

PR #333, opened 2025-10-10
Add ModelConfig.md model config documentation

favyen2:


docs/ModelConfig.md
```diff
@@ -0,0 +1,748 @@
...
+      # ... initial weight configuration
+    print_parameters: false
+    print_model: false
+```
+
+### model
+
+The model argument is a `torch.nn.Module` that corresponds to the model architecture.
+Typically, models will either employ SingleTaskModel (when training a model for one
+task, like segmentation or detection) or MultiTaskModel (when training a model on
```
APatrickJ:
```suggestion
task, like segmentation or detection) or `MultiTaskModel` (when training a model on
```

docs/ModelConfig.md
```diff
@@ -0,0 +1,748 @@
...
+multiple tasks, e.g. predicting LFMC at each pixel while also classifying land cover at
+each pixel). These classes provide scaffolding that makes foundation models like
+OlmoEarth and SatlasPretrain compatible with decoders like U-Net and Faster R-CNN.
+
+Here is an example with SingleTaskModel, for segmentation:
```
APatrickJ:
```suggestion
Here is an example with `SingleTaskModel`, for segmentation:
```

docs/ModelConfig.md
```diff
@@ -0,0 +1,748 @@
...
+trainer:
+  # Lightning trainer options and callbacks.
+```
+
+The YAML is parsed by jsonargparse, so each section directly configures a Python class
+or other type, and you can check those classes in the rslearn codebase for more
+details.
+
+Below, we dive into each of these components. This documentation is intended to
+supplement the examples in [Examples](Examples.md).
```
APatrickJ: Not an issue for this PR, but only one of the examples linked there actually works. And the actual model training part hasn't been written yet. For the future, I think it would be useful to have a couple of practical examples to show a full model config end-to-end, maybe one for segmentation and one for regression, so the reader can follow along with the examples as they read through each of these sections.

docs/ModelConfig.md
```diff
@@ -0,0 +1,748 @@
...
+        # Either one or three bands should be specified. The image is expected to be
+        # under the "image" key.
+        image_bands: [0, 1, 2]
+        # This defines an optional linear scaling from the raw image values to pixel
+        # values that should be saved. This is generally needed since usually the input
+        # image to the model will have been normalized already.
+        remap_values: [[0, 1], [0, 255]]
+```
+
+For example, suppose you have a DataInput that loads all Sentinel-2 L2A bands:
```
APatrickJ:
```suggestion
For example, suppose you have a `DataInput` that loads all Sentinel-2 L2A bands:
```

docs/ModelConfig.md
```diff
@@ -0,0 +1,748 @@
...
+            segment_input: "targets
+```
+
+### num_workers and init_workers
+
+`num_workers` controls the number of data loader workers.
+
+Worker processes are also used to enumerate the windows in the rslearn dataset, and to
+determine which ones contain the layers that are needed. By default, the same number of
+workers used for data loader is used for this initialization. However, it can be
```
APatrickJ:
```suggestion
workers used for the data loader is used for this initialization. However, it can be
```

APatrickJ (approved):
Thanks for writing this up! I pointed out a few minor typos and small ways to clarify. I think in the future when the examples section is expanded, we could tie the examples in more with this doc so the reader can follow a real example as they read through each section. But as is, it looks good to me.

PR #335, opened 2025-10-15
Resolves pystac memleak (hopefully)

cmwilhelm:
Our Planetary Computer data sources use the [pystac](https://pystac.readthedocs.io/en/stable/index.html)
library as an HTTP client to retrieve and normalize data
from Microsoft's API.

This library is designed to build up object graphs / caches in-memory
over the lifetime of a process and does not appear to be intended
for use in long-running applications. There is no way to opt out of,
cap the growth of, or clear the cache. Every pystac object, including
the client, is registered as a node in the object graph, and that
graph gets added to whenever results resolve from HTTP responses.

The consequence is that when running dataset handler operations
through our planetary computer DSes, we are experiencing unbounded
memory growth. In my local tests, this could be as much as 100MB/min
per worker process. Previously we worked around this by running handler
operations in a multiproc pool with limited executions per worker process.

This changeset instead sets a max-Items budget for a pystac `Client`
object. When this budget is exceeded, the Client is replaced with a fresh
client. Due to the stickiness of the multi-way references on pystac graph nodes,
I found it was necessary to do two additional things to make the client and its friends
eligible for garbage collection:

* clear members off the Client node (`root.clear_links()` etc)
* stop using `Collections` and `CollectionClients`

On the latter -- the collection client we were using previously was mostly
sugar around an HTTP request we can perform directly off the parent
client in `get_by_name`. Note this will use the search/ endpoint instead
of the collection/ endpoint now, but according to
https://planetarycomputer.microsoft.com/api/stac/v1
planetary computer is conformant to both.

Note, this is a breaking change specifically for the satlas data sources in RSLP,
which access but do not use the collection client previously initialized by
the Sentinel1/2 datasources. This will require a tiny update in rslp code, accordingly.

Memory growth before change:
Screenshot 2025-10-15 at 12 58 09 PM

Memory growth after change:
Screenshot 2025-10-15 at 12 57 31 PM

Note initial growth before GC has run enough times to start
ejecting long survivors, after which we reach equilibrium.


rslearn/data_sources/planetary_computer.py
```diff
@@ -151,20 +167,40 @@ def from_config(config: RasterLayerConfig, ds_path: UPath) -> "PlanetaryComputer
 
     def _load_client(
         self,
-    ) -> tuple[pystac_client.Client, pystac_client.CollectionClient]:
+    ) -> pystac_client.Client:
```
cmwilhelm: This interface change is the breaking part that will require us/me to modify the `Monthly*` datasources in rslp.

PR #339, opened 2025-10-18
[DRAFT PROPOSAL DO NOT MERGE] draft pystac replacement

cmwilhelm:
In a previous PR I tried to address a memory leak originating in
the pystac/pystac_client libraries. We use these libraries to retrieve
data from Planetary Computer (and earthdaily, actually), which implements
the STAC standard.

As discussed in that PR, the memory "leak" is due to the library maintaining
a huge object graph/cache to aid in catalog and collection traversal over the lifetime
of a process. You can't opt out of this or bound it. It's also something we
don't need. My workaround was to kill and restart the client periodically,
and try to wipe all its entangled objects so they can be garbage collected.

This seems to work, but I have a few thoughts:

1. The memory footprint is still pretty big because the GC takes a while
  to be able to clear those circular object graphs.
2. Our dataset build nodes in Run are super underutilizing CPU. We could
  easily spawn tons more worker processes -- IF we can handle the memory footprint.
3. The way we use the pystac client is EXTREMELY light.

On (3) in particular-- we are only doing some very simple searches to retrieve
items and assets, meaning the thousands and thousands of lines of catalog
traversal code (and data structures) go unused. All the endpoints we hit are STAC 1.0,
meaning all the complicated normalization + data migration code that is implemented
in pystac item deserializers is unnecessary. As far as the in-memory cache goes --
our data source classes are actually caching the data they need to disk and short-circuiting
calls through to the client anyway!

All that's really left is a pretty simple HTTP client that can handle structured search
over paginated results, and then deserialize the responses into something familiar
our data sources can deal with.

This PR is just a draft. I asked Cursor to lift the salient parts out of the pystac
client and reduce to the bare essentials. I'm wondering what we think about getting
rid of the dependency on pystac/pystac-client and just maintaining this (relatively)
small utility file of our own. It does only what we need to do, doesn't have the memleak,
would be easy for us to instrument ourselves, etc. There's obviously a maintenance
burden here that we outsource with the use of a 3rd party client lib, but I'm offering
this as food for thought.

### DIFF
--- a/rslearn/data_sources/planetary_computer.py
+++ b/rslearn/data_sources/planetary_computer.py
@@ -10,8 +10,6 @@ from typing import Any
 import affine
 import numpy.typing as npt
 import planetary_computer
-import pystac
-import pystac_client
 import rasterio
 import requests
 import shapely
@@ -21,6 +19,7 @@ from upath import UPath
 from rslearn.config import LayerConfig, QueryConfig, RasterLayerConfig
 from rslearn.const import WGS84_PROJECTION
 from rslearn.data_sources import DataSource, Item
+from rslearn.data_sources.stac_utils import StacApiClient
 from rslearn.data_sources.utils import match_candidate_items_to_window
 from rslearn.dataset import Window
 from rslearn.dataset.materialize import RasterMaterializer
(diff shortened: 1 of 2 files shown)


favyen2:
Closing since I guess we ended up doing something similar in a different PR, it doesn't have the retries though which might be useful (it makes sense for data sources that frequently encounter transient errors to do some retries within the data source).

PR #365, opened 2025-11-12
Add resizing to TerraMind

yawenzzzz:


rslearn/models/terramind.py
```diff
@@ -132,6 +138,15 @@ def forward(self, inputs: list[dict[str, Any]]) -> list[torch.Tensor]:
             if modality not in inputs[0]:
                 continue
             cur = torch.stack([inp[modality] for inp in inputs], dim=0)  # (B, C, H, W)
+            if self.do_resizing and (
```
Hgherzog: What about for single pixel time series? We should resize to the base patch size imo to be consistent with what we have done

yawenzzzz: OK sounds good!

Hgherzog (approved):

PR #371, opened 2025-11-18
Improve dataset configuration

favyen2:
- Switch from manual parsing to pydantic models for the bulk of the dataset config.
- Switch from manual parsing to jsonargparse for initializing data sources. Now, the dataset config has class_path and init_args fields for the data source, where the latter is a dict handled by jsonargparse. Some backwards compatibility is maintained.
- Add DataSourceContext to pass the dataset path and LayerConfig to the data source. Most data sources use the context to do things like adjust file paths that are relative to the dataset root directory. The context is passed to the data source by injecting it into the init_args. Not the cleanest solution, but it seemed better than passing it via a method after instantiation.
- Remove the RasterFormats and VectorFormats class registries, and the manual argument parsing. Instead, these are now also initialized via jsonargparse, and the class_path is directly set from the dataset config. Some backwards compatibility for RasterFormats is maintained.
- Remove the Materializers registry. We have been only using RasterMaterializer and VectorMaterializer for quite some time, so now we just directly initialize them depending on the layer type.
- Remove rslearn.data_sources.raster_source. This module provided is_raster_needed, but that was only still used in gcp_public_data, and I have changed gcp_public_data now to determine the needed assets upon initialization (similar to other data sources).
- Remove TileStore configuration backwards compatibility. Now only the jsonargparse format is accepted. This shouldn't break much because we rarely specify the tile_store in the dataset config, and the old format has been deprecated for a while (although it looks like it was still copied and pasted recently into a few places, mostly in tests which I have updated).

### DIFF
--- a/tests/unit/config/test_dataset.py
+++ b/tests/unit/config/test_dataset.py
@@ -243,3 +243,23 @@ class TestBackwardsCompatibility:
         assert ds.query["sar:instrument_mode"] == {"eq": "IW"}
         assert ds.cache_dir is not None
         assert ds.cache_dir.path == "cache/planetary_computer"
+
+    def test_raster_format_compat(self) -> None:
+        """Check parsing for legacy raster format config format."""
+        layer_config = LayerConfig.model_validate(
+            {
+                "band_sets": [
+                    {
+                        "bands": ["mask"],
+                        "dtype": "uint8",
+                        "format": {"format": "png", "name": "single_image"},
+                    }
+                ],
+                "type": "raster",
+            }
+        )
+        assert len(layer_config.band_sets) == 1
+        assert layer_config.band_sets[0].bands == ["mask"]
+        raster_format = layer_config.band_sets[0].instantiate_raster_format()
+        assert isinstance(raster_format, SingleImageRasterFormat)
+        assert raster_format.format == "png"


cmwilhelm:
@favyen2 can you elaborate on the scope of the changes you're referring to here?

> This change will require modifications to all existing dataset configs.

I haven't looked at this PR deeply; the summary notes seem positive. Still, I'm wondering if changes that break core API contracts should be presented in design docs to the broader group at this point.

favyen2:
Here is an example.

Old version:

```json
    "sentinel2": {
      "band_sets": [{
          "bands": ["B01", "B02", "B03", "B04", "B05", "B06", "B07", "B08", "B8A", "B09", "B11", "B12"],
          "dtype": "uint16"
      }],
      "data_source": {
        "cache_dir": "cache/planetary_computer",
        "duration": "270d",
        "harmonize": true,
        "ingest": false,
        "name": "rslearn.data_sources.planetary_computer.Sentinel2",
        "query_config": {
          "max_matches": 6,
          "min_matches": 6,
          "period_duration": "30d",
          "space_mode": "PER_PERIOD_MOSAIC"
        },
        "sort_by": "eo:cloud_cover",
        "time_offset": "-90d"
      },
      "type": "raster"
    }
```

New version:
```json
    "sentinel2": {
      "band_sets": [{
          "bands": ["B01", "B02", "B03", "B04", "B05", "B06", "B07", "B08", "B8A", "B09", "B11", "B12"],
          "dtype": "uint16"
      }],
      "data_source": {
        "class_path": "rslearn.data_sources.planetary_computer.Sentinel2",
        "init_args": {
          "cache_dir": "cache/planetary_computer",
          "harmonize": true,
          "sort_by": "eo:cloud_cover",
        },
        "duration": "270d",
        "time_offset": "-90d",
        "ingest": false,
        "query_config": {
          "max_matches": 6,
          "min_matches": 6,
          "period_duration": "30d",
          "space_mode": "per_period_mosaic"
        }
      },
      "layer_type": "raster"
    }
```

We can change things to avoid the "type" -> "layer_type" and "PER_PERIOD_MOSAIC" -> "per_period_mosaic" changes, but the separation of generic data source configuration options (like duration, time_offset, and query_config) from source-specific ones (like cache_dir, harmonize, and sort_by that are arguments to rslearn.data_sources.planetary_computer.Sentinel2) is sort of integral to this change (otherwise there isn't a good way to e.g. throw an error if an unknown key is passed, because different parts of the system won't know if there are extra config options that will be read from the same config section by other parts of the system).

rslearn/data_sources/planetary_computer.py
```diff
@@ -581,72 +559,45 @@ class Sentinel2(PlanetaryComputer):
...
-            asset_bands = self.BANDS
-        else:
+        # Determine which assets we need based on the bands in the layer config.
+        if context.layer_config is not None:
+            asset_bands: dict[str, list[str]] = {}
+            for asset_key, band_names in self.BANDS.items():
+                for band_set in context.layer_config.band_sets:
+                    if not set(band_set.bands).intersection(set(band_names)):
+                        continue
+                    asset_bands[asset_key] = band_names
```
APatrickJ: Should we add a `break` here to exit the inner loop once a match is found?

rslearn/data_sources/usgs_landsat.py
APatrickJ: Remove this

APatrickJ (commented):
Thanks for taking this on!

rslearn/data_sources/planetary_computer.py
```diff
@@ -581,72 +559,45 @@ class Sentinel2(PlanetaryComputer):
...
+        if assets is not None:
             asset_bands = {asset_key: self.BANDS[asset_key] for asset_key in assets}
```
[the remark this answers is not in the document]
favyen2: ah thanks, fixed!

rslearn/config/dataset.py
```diff
@@ -332,6 +392,50 @@ class DataSourceConfig(BaseModel):
         description="Whether to ingest this layer (default True). If False, it will be directly materialized without ingestion.",
     )
 
+    @model_validator(mode="before")
+    @classmethod
+    def convert_from_legacy(cls, d: dict[str, Any]) -> dict[str, Any]:
```
APatrickJ: We may want to add some tests around this behavior, suggest putting them into a separate test class so they are isolating and can be easily deleted later once we remove the legacy backward compatibility code

PR #373, opened 2025-11-20
Support concatenating features

yawenzzzz:
This is a simple approach for late fusion, it supports concatenating embedding from OlmoEarth with additional features (like era5 time-series). Here's an example of using this: https://github.com/allenai/rslearn_projects/blob/yawenz/20251118_awf_add_features/data/helios/v2_awf_lulc/finetune_s2_20251118.yaml.

### DIFF
--- a/rslearn/models/concatenate_features.py
+++ b/rslearn/models/concatenate_features.py
@@ -16,6 +16,7 @@ class ConcatenateFeatures(torch.nn.Module):
         out_channels: int | None = None,
         num_conv_layers: int = 1,
         kernel_size: int = 3,
+        final_relu: bool = False,
     ):
         """Create a new ConcatenateFeatures.
 
@@ -26,6 +27,7 @@ class ConcatenateFeatures(torch.nn.Module):
             out_channels: number of output channels of the additional features.
             num_conv_layers: number of convolutional layers to apply to the additional features.
             kernel_size: kernel size of the convolutional layers.
+            final_relu: whether to apply a ReLU activation to the final output, default False.
         """
         super().__init__()
         self.key = key
@@ -48,7 +50,7 @@ class ConcatenateFeatures(torch.nn.Module):
                     padding="same",
                 )
             )
-            if i < num_conv_layers - 1:
+            if i < num_conv_layers - 1 or final_relu:
                 conv_layers.append(torch.nn.ReLU(inplace=True))
 
         self.conv_layers = torch.nn.Sequential(*conv_layers)


rslearn/models/concatenate_features.py
```diff
@@ -0,0 +1,60 @@
...
+        super().__init__()
+        self.key = key
+
+    def forward(
+        self, features: list[torch.Tensor], inputs: list[dict[str, Any]]
+    ) -> list[torch.Tensor]:
+        """Concatenate the feature map with the raw data inputs.
+
+        Args:
+            features: list of feature maps at different resolutions.
```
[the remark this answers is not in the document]
Farbum: Reading the doc, this is useful.

Yes that's exactly what I was expecting as a first naive read, B being part of the tensor, as opposed to be repeated B times. I still think that adding an "assert" to match batch size could be a good add (in addition to the doc) for this concatenate module, as it greatly improves readability, and the code will break anyway if there is no match.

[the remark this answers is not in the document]
Farbum: Love that!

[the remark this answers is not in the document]
yawenzzzz: Added support for multi-resolution feature map and also conv layers.

tests/unit/models/test_concatenate_features.py
```diff
@@ -0,0 +1,32 @@
+"""Tests for the rslearn.models.concatenate_features module."""
+
+import torch
+
+from rslearn.models.concatenate_features import ConcatenateFeatures
```
favyen2: may be worth having test that runs with one conv

rslearn/models/concatenate_features.py
```diff
@@ -0,0 +1,81 @@
...
+        super().__init__()
+        self.key = key
+
+        conv_layers = []
+        for _ in range(num_conv_layers):
+            conv_layers.extend(
+                [
+                    torch.nn.Conv2d(
+                        in_channels=in_channels,
+                        out_channels=out_channels,
```
favyen2: I think this won't work if you have more than one conv layer if in_channels != out_channels.

tests/unit/models/test_concatenate_features.py
favyen2: Oh I guess it does test with one conv. In that case might be worth having test that runs with zero conv.

rslearn/models/concatenate_features.py
yawenzzzz: Fixed! thanks for identifying the bug!

PR #376, opened 2025-11-26
Make subcommand optional

favyen2:
This is to support running RslearnLightningCLI(..., run=False) to instantiate the trainer and model without running an actual Lightning stage.

WIP because there may be a better way to do this. The main goal for now is to be able to easily instantiate a model and save its checkpoint.


favyen2:
It now seems like this is useful after all -- it is good for testing the overall jsonargparse testing, and @gabrieltseng also had a case where he wanted to initialize data module from config.

PR #377, opened 2025-12-01
Update OlmoEarth model wrapper to raise error if checkpoint doesn't exist

favyen2:
Update OlmoEarth model wrapper to raise error if checkpoint doesn't exist.

It originally raised error but looks like this got lost after some re-shuffling of functionality for loading from previously fine-tuned checkpoints (this is now done by using RestoreConfig).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "rslearn"
-version = "0.0.16"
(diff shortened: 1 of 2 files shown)


cmwilhelm (approved):
LGTM. Can you bump the rslearn version and kick off a release so we can target this update in olmoearth_run?

PR #379, opened 2025-12-02
Model component API improvements

favyen2:
Model component API improvements.

- Allow intermediate outputs (passed between model components) to be arbitrary, but provide standard types that can be used in most scenarios. This makes it more clear what the expected input and output for each model component is.
- Make the encoder components (after the initial feature extractor) and decoder components (before the final predictor) share the same API.
- Create a new SampleMetadata class that replaces the metadata dict.
- Create a new ModelContext class that includes the input dicts and SampleMetadatas to pass to models instead of just the inputs. This way models can use location/timestamp info if it's useful.
- Add abstract base classes for FeatureExtractor, IntermediateComponent, and Predictor so the signature of the forward pass is clear for all these model components.
- Improve the documentation and type checking for task-specific outputs (output from Predictor that gets split up per-example and eventually passed to Task.process_output).

Potential config-breaking changes:
- I removed the `collapse` option from PickFeatures. This was previously used to turn a list of BCHW feature maps into one BCHW tensor, since many task heads expect one BCHW tensor as input. But now those heads expect a `FeatureMaps` instead so `collapse` is not needed.

### DIFF
--- a/rslearn/models/singletask.py
+++ b/rslearn/models/singletask.py
@@ -46,7 +46,6 @@ class SingleTaskModel(torch.nn.Module):
         Args:
             context: the model context.
             targets: optional list of target dicts
-            info: optional dictionary of info to pass to the last module
 
         Returns:
             the model output.


rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -196,7 +202,7 @@ def forward(self, inputs: list[dict[str, Any]]) -> list[torch.Tensor]:
         # Note that only months (0 to 11) are used in OlmoEarth position encoding.
         # For now, we assign same timestamps to all inputs, but later we should handle varying timestamps per input.
         timestamps = torch.zeros(
-            (len(inputs), max_timesteps, 3), dtype=torch.int32, device=device
+            (len(context.inputs), max_timesteps, 3), dtype=torch.int32, device=device
         )
```
yawenzzzz: We may want to update this part using the actual `time_range`, but it will require using the `max_timesteps` to figure out the time interval.

favyen2: Hm yeah originally I wasn't thinking to actually change the models yet in this PR, but maybe it's worth it given that the information currently passed may not be sufficient.

Do you think we should compute a time range based on the min/max timestamp of Items in the corresponding WindowLayerData (from items.json)? And pass that to the model somehow.

yawenzzzz: yes, that's right, the item time range should be more accurate than the window time range.

PR #381, opened 2025-12-04
Implement metadata storage abstraction for rslearn datasets

favyen2:
Implement metadata storage abstraction for rslearn datasets.

This enables metadata to be stored using methods other than the current file-based storage, such as storing metadata in a database.

- DatasetConfig updated to include a StorageConfig that configures a WindowStorageFactory.
- WindowStorageFactory provides a method that creates a WindowStorage from a dataset path (so that we don't need the dataset path in the StorageConfig).
- The WindowStorage various functions to get and create windows, get and update window layer datas, and get and update completed layers in windows.
- I removed the dataset index since we haven't been using it and this new WindowStorage abstraction enables better options (like storing metadata in a database) to increase metadata access speed.

Notes:
- Dataset configs should not require any updates, but the `Window.__init__` has been updated to take a WindowStorage instead of a UPath, so tools calling that function will need to be updated.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "rslearn"
-version = "0.0.17"
+version = "0.0.18"
 description = "A library for developing remote sensing datasets and models"
 authors = [
     { name = "OlmoEarth Team" },
--- a/uv.lock
+++ b/uv.lock
@@ -1,5 +1,5 @@
 version = 1
-revision = 2
+revision = 3
 requires-python = ">=3.11"
 resolution-markers = [
     "python_full_version >= '3.14' and platform_machine == 'aarch64' and sys_platform == 'linux'",
@@ -4320,7 +4320,7 @@ wheels = [
 
 [[package]]
 name = "rslearn"
-version = "0.0.17"
+version = "0.0.18"
 source = { editable = "." }
 dependencies = [
     { name = "boto3" },


rslearn/dataset/storage/file.py
```diff
@@ -0,0 +1,248 @@
...
+    Returns:
+        the window object.
+    """
+    metadata_fname = window_dir / "metadata.json"
+    with metadata_fname.open() as f:
+        metadata = json.load(f)
+    return Window.from_metadata(storage, metadata)
+
+
+class FileDatasetStorage(DatasetStorage):
```
APatrickJ: Do we also need this to understand `bands.json`?

https://github.com/allenai/rslearn/blob/2cbb8665169e12e0b9c0e59d705f94b3a1862bfb/rslearn/tile_stores/default.py#L39-L40

rslearn/dataset/storage/storage.py
```diff
@@ -0,0 +1,137 @@
...
+    This is instantiated by a DatasetStorageFactory.
+    """
+
+    @abc.abstractmethod
+    def get_window_root(self, group: str, name: str) -> UPath:
+        """Get the path where the window should be stored."""
+        raise NotImplementedError
+
+    @abc.abstractmethod
+    def get_windows(
```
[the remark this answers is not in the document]
favyen2: We could call it WindowDatabase? So WindowDatabaseFactory, WindowDatabase, and FileWindowDatabaseFactory/FileWindowDatabase.

rslearn/dataset/window.py
```diff
@@ -1,19 +1,18 @@
...
 
+from rslearn.dataset.storage.storage import DatasetStorage
 from rslearn.log_utils import get_logger
 from rslearn.utils import Projection, STGeometry
-from rslearn.utils.fsspec import open_atomic
 from rslearn.utils.raster_format import get_bandset_dirname
 
 if TYPE_CHECKING:
-    from .index import DatasetIndex
+    pass
```
APatrickJ: Should be able to remove this entirely

rslearn/dataset/storage/storage.py
```diff
@@ -0,0 +1,137 @@
...
+        self,
+        groups: list[str] | None = None,
+        names: list[str] | None = None,
+        show_progress: bool = False,
+        workers: int = 0,
```
APatrickJ: It seems like these could be constructor arguments. I'm not sure a database implementation will want to have them.

favyen2: Yeah I wasn't sure what to do about these arguments, they don't fit here but I didn't see anywhere else to put them either.

They don't quite make sense to me as constructor arguments from the dataset config because they would depend on the environment and such -- currently these get passed to here via the `--workers` argument to `rslearn dataset prepare/ingest/materialize`. They could be keyword arguments but then I'm not sure how they would get passed here. Or they could be somehow passed via environment variable? That doesn't seem great either.

Actually we could define these extra optional arguments on `FileWindowStorage` only, and in `main.py` we can see if it is `isinstance(dataset.storage, FileWindowStorage)` and pass those command-line options over only if it is, how does that sound? Or do you have other ideas?

rslearn/dataset/storage/file.py
```diff
@@ -0,0 +1,248 @@
...
+        workers: int = 0,
+    ) -> list["Window"]:
+        """Load the windows in the dataset.
+
+        Args:
+            groups: an optional list of groups to filter loading
+            names: an optional list of window names to filter loading
+            show_progress: whether to show tqdm progress bar
+            workers: number of parallel workers, default 0 (use main thread only to load windows)
+        """
```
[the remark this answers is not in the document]
favyen2: Actually there is no `@override` in Python 3.11, do you have any suggestions about it?

rslearn/dataset/storage/storage.py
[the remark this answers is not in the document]
favyen2: Sounds good to me

rslearn/dataset/storage/storage.py
APatrickJ: Something like this maybe?

```python
from typing import Unpack

class FileWindowLoadOptions(TypedDict, total=False):
    show_progress: bool
    workers: int

class FileWindowStorage(WindowStorage):
    """The default file-backed window storage."""

    def __init__(self, path: UPath):
        self.path = path


    @override
    def get_windows(
        self,
        groups: list[str] | None = None,
        names: list[str] | None = None,
        **options: Unpack[FileWindowLoadOptions],
    ) -> list["Window"]:
        show_progress = options.get("show_progress", False)
        workers = options.get("workers", 0)
```

And for the Dataset class impl:

```python
    def load_windows(
        self,
        groups: list[str] | None = None,
        names: list[str] | None = None,
        **options: Any,
    ) -> list[Window]:
        """Load the windows in the dataset.

            names: an optional list of window names to filter loading
            show_progress: whether to show tqdm progress bar
            workers: number of parallel workers, default 0 (use main thread only to load windows)
        """
        return self.storage.get_windows(
            groups=groups, names=names, **options
        )
```

kwargs certainly have their drawbacks, but this might be an appropriate use case for them. And at least there's some attempt at typing checking.

PR #382, opened 2025-12-05
Update RslearnWriter to respect the configured BandSetConfig.dtype

favyen2:
Update RslearnWriter to respect the configured BandSetConfig.dtype.

- RasterMerger now initializes the merged output to the configured dtype instead of copying it from the first PendingPatchOutput.
- Casting to uint8 removed from SegmentationTask.process_output.
- Update documentation of BandSetConfig.dtype to make it clear that it's only used for dataset materialize and (now) model predict.
- Add test to make sure the casting works.


cmwilhelm (approved):
It looks nice to me, but I did have the one clarifying question. Thanks for being so fast about everything Favyen.

PR #384, opened 2025-12-10
Update model component docs to reflect the updated model component API.

favyen2:


docs/TasksAndModels.md
```diff
@@ -90,8 +93,15 @@ The configuration snippet below summarizes the most common options. See
         positive_class_threshold: 0.75 # predict as cls_name if corresponding probability exceeds this threshold
 ```
```
Farbum: Would it make sense, for the sake of completeness, to include that the `layers` and `bands` parameters of the layer inputs have to match the data config file?
This is something that was not immediately obvious when I first tried to fine-tune a model.

Farbum (commented):
LGTM, I left a few minor comments.

Since certain predictors (and maybe certain intermediate components) require specific shapes (e.g. segmentation head requires predicted classes at dimension 1) ), I think it would be useful to reflect that in the specific predictor (or if/where applicable in the abstract base class) doc-strings.
If that sounds good to you, we can update them as we go.

docs/TasksAndModels.md
favyen2: This is documented in https://github.com/allenai/rslearn/blob/master/docs/ModelConfig.md#inputs so I updated TasksAndModels.md to have a link back to there.

Farbum (approved):

PR #385, opened 2025-12-11
Fix SimpleTimeSeries so it passes ModelContext to the feature extractor.

favyen2:
It was broken with the model component API update since it was not updated to pass a ModelContext instead of the input dicts directly.

I added a test so it will catch it in the future.


Hgherzog (approved):
LGTM

PR #386, opened 2025-12-11
Fix bug with AllPatchesDataset when the window is smaller than the patch size

favyen2:
Previously, the dataset would pad the image data by patch size on the right/bottom, so that if the patch bounds exceed the window bounds then it's okay. However, when computing patches, it would previously align with the right/bottom of the window, and extend past the left/top of the window; so if the window is (0, 0, 2, 2) with patch size 4, it would return the patch (-2, -2, 2, 2), but the image data wasn't padded to support that access.

The result is an image with width/height 0 instead of width/height matching the patch size.

Now, instead, when computing patches, we ensure it returns (0, 0, 4, 4) for that case, which works because of the padding to the right/bottom.

I added a test for this too.

### DIFF
--- a/rslearn/arg_parser.py
+++ b/rslearn/arg_parser.py
@@ -20,6 +20,7 @@ class RslearnArgumentParser(LightningArgumentParser):
     def parse_string(
         self,
         cfg_str: str,
+        *args: Any,
         **kwargs: Any,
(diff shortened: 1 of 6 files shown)


yawenzzzz (approved):
LGTM!

PR #393, opened 2025-12-16
Hadriens/model seq len

Farbum:
Problem: current behavior does not allow for variable temporal sequence lengths across samples within batch for a given modality, and it does not allow for variable temporal sequence lengths across modalities.
This addresses the issue by:
    - Scanning input tensors to find the maximum temporal sequence length across samples and modalities
    - Pad all input tensors (across modalities) to the same sequence length
    
This is still not ideal as the padding happens at the end of the sequence, e.g.
A valid tensor of shape [H W t C] with t < max_t will be padded to [H W max_t, C] which means all invalid (masked) time steps are at the end of the temporal sequence.

Future step should be to infer from the window metadata (or any other mechanism) which time steps are missing to properly impute missing values at the right place in the sequence.

### DIFF
--- a/rslearn/models/olmoearth_pretrain/model.py
+++ b/rslearn/models/olmoearth_pretrain/model.py
@@ -258,7 +258,7 @@ class OlmoEarth(FeatureExtractor):
 
         return MaskedOlmoEarthSample(**kwargs), present_modalities, device
 
-    def forward(self, context: ModelContext) -> FeatureMaps:
+    def forward(self, context: ModelContext) -> FeatureMaps | TokenFeatureMaps:
         """Compute feature maps from the OlmoEarth backbone.
 
         Args:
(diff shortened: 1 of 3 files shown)


rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -165,49 +160,87 @@ def _load_model_from_checkpoint(
...
-            # Create mask array which is BHWTS (without channels but with band sets).
-            num_band_sets = len(Modality.get(modality).band_sets)
-            mask_shape = cur.shape[0:4] + (num_band_sets,)
-            mask = (
-                torch.ones(mask_shape, dtype=torch.int32, device=device)
-                * MaskValue.ONLINE_ENCODER.value
+
+            # Create mask: ONLINE_ENCODER for valid, MISSING for padded timesteps
+            b, h, w = cur.shape[0], cur.shape[1], cur.shape[2]
+            mask = torch.full(
```
favyen2: Since you are setting a mask now, fast_pass should not be used since that will tell olmoearth_pretrain to ignore the mask. I think you'll need to update that to pass fast_pass only if the mask is not all ONLINE_ENCODER.

I think it's still important to set fast_pass if nothing is masked, because it increases the speed by about 2x, and for most applications so far we don't have missing timesteps (or at least ignore examples with missing timesteps).

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -165,49 +160,87 @@ def _load_model_from_checkpoint(
...
+                (b, h, w, max_timesteps, num_band_sets),
+                fill_value=MaskValue.ONLINE_ENCODER.value,
+                dtype=torch.int32,
+                device=device,
             )
+            for i, orig_t in enumerate(original_timesteps):
```
favyen2: Suggest to use `sample_idx` or `example_idx` instead of `i`.

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -165,49 +160,87 @@ def _load_model_from_checkpoint(
...
-                an additional N dimension for the tokens.
+            tuple of (sample, present_modalities, device)
         """
         kwargs = {}
         present_modalities = []
         device = None
-        # Handle the case where some modalities are multitemporal and some are not.
-        # We assume all multitemporal modalities have the same number of timesteps.
+
+        # First pass: find global max_timesteps across all modalities
```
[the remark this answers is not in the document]
Farbum: Done

PR #395, opened 2025-12-16
Add hourly ERA5

yawenzzzz:
Added ERA5 hourly dataset, as discussed, the API request will get all the hourly data within a month and then stack them together, therefore, the first dimension of the geotiff is timesteps * num_variables. For example, if we request January, timesteps = 744 (31 x 24). 

Also, instead of always getting global data (which is fine for monthly average data), we can specify the bounds (min_lon, min_lat, max_lon, max_lat), which can help speed up the data ingestion for hourly data. As a reference, it took 18min to get one month of global data (the nc file is ~4.8GB), and 3min to get 1x1 degree data (the nc file is ~86K).

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "rslearn"
-version = "0.0.18"
+version = "0.0.20"
 description = "A library for developing remote sensing datasets and models"
 authors = [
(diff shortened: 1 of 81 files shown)


Farbum (approved):
LGTM, thanks Yawen!

PR #397, opened 2025-12-17
Fix jsonargparse parsing again

favyen2:
Fix jsonargparse parsing again.
    
It was broken because in RslearnArgumentParser we took **kwargs but not *args and sometimes jsonparparse passes positional instead of keyword arguments.
    
I added a test for it that was failing with jsonargparse 4.35.0. With the change it is working. I also updated uv.lock and with 4.44.0 it seems to work with or without the change, but anyway better to update it for future-proofing.

This PR depends on #376


Farbum (approved):
LGTM

PR #398, opened 2025-12-18
Pass dates through to the model

gabrieltseng:
This PR introduces a `RasterImage` dataclass, which replaces the `torch.Tensor` usually used to represent raster images.

A `RasterImage` also contains an optional `timestamps` list which records the timestamp associated with each image. This is pulled from the `items.json`.

The main change is relatively contained (in [train/dataset.py](https://github.com/allenai/rslearn/blob/25c983ce01cc1869a30d92264616a46bbaa86e03/rslearn/train/dataset.py)), but all tasks, transforms and models also need to be updated to handle this change.

So far, this change has been implemented for OlmoEarth and the subset of transforms required for the [Mozambique LULC project](https://github.com/allenai/olmoearth_projects/tree/main/olmoearth_run_data/mozambique_lulc). An example run with this new approach is [here](https://wandb.ai/eai-ai2/2025_09_18_mozambique_lulc/runs/xhydwwoe). 

Todo:
- [x] Update all tasks
- [x] Successful run: https://wandb.ai/eai-ai2/2025_09_18_mozambique_lulc/runs/xhydwwoe
- [x] Update all models
    - [x] Molmo 
    - [x] CLIP
    - [x] DINOv3
    - [x] CROMA
    - [x] OlmoEarth
    - [x] Faster RCNN     
    - [x] Panopticon
    - [x] Prithvi 
    - [x] Terramind 
    - [x] Swin 
    - [x] SSL4EO 
    - [x] Anysat
    - [x] Panopticon
    - [x] Presto
    - [x] Simple time series
    - [x] Satlas
    - [x] Galileo
    - [x] SAM2 
- [x] Update all transforms
- [x] All tests passing


rslearn/train/all_patches_dataset.py
```diff
@@ -294,6 +295,30 @@ def crop_input_dict(d: dict[str, Any]) -> dict[str, Any]:
                                     scaled_start[1] : scaled_end[1],
                                     scaled_start[0] : scaled_end[0],
                                 ].clone()
+                            elif isinstance(value, RasterImage):
```
favyen2: Might be able to reuse some helper function here instead of repeating the logic twice.

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -165,111 +166,94 @@ def _load_model_from_checkpoint(
...
-        """Prepare modality tensors and masks for the OlmoEarth model.
+    @staticmethod
+    def time_ranges_to_timestamps(
+        time_ranges: list[tuple[datetime, datetime]], device: torch.device
+    ) -> torch.Tensor:
+        """Turn the time ranges stored in a RasterImage to timestamps accepted by OlmoEarth."""
+        timestamps = torch.zeros(
+            (len(time_ranges), 3), dtype=torch.int32, device=device
+        )
+        mid_ranges = [t[0] + ((t[1] - t[0]) / 2) for t in time_ranges]
```
favyen2: It would be useful to have a comment here saying that the images have time ranges but we only pass year/month/day (with only month being used AFAIK) to olmoearth_pretrain, so we opt to take the middle of the time range.

rslearn/utils/raster_format.py
```diff
@@ -22,6 +25,21 @@
 logger = get_logger(__name__)
 
 
+@dataclass
+class RasterImage:
+    """A raster image is a torch.tensor containing the images and their associated timestamps."""
+
+    # image will have an extra temporal dimension, CTHW
+    # TODO - should we collapse if it is only one timestamp?
```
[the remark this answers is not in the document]
gabrieltseng: removed the TODO

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -165,111 +166,94 @@ def _load_model_from_checkpoint(
...
-
-        # Second pass: pad and process each modality with global max_timesteps
-        for modality in present_modalities:
-            tensors, num_bands, num_band_sets = modality_data[modality]
-            target_ch = max_timesteps * num_bands
-
-            # Pad tensors to target_ch and track original timesteps for masking
-            padded = []
-            original_timesteps = []
-            for t in tensors:
```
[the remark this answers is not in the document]
gabrieltseng: yes this makes sense. I added a `use_legacy_timestamps` initialization argument to do this

rslearn/utils/raster_format.py
```diff
@@ -22,6 +25,29 @@
 logger = get_logger(__name__)
 
 
+@dataclass
```
favyen2: This is training-specific so I think it fits better in `rslearn/train/model_context.py`.

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -242,19 +286,31 @@ def _prepare_modality_inputs(
...
+            if max([len(t) for t in timestamps_per_instance]) == 0:
+                # Timestamps is required.
+                raise ValueError("No inputs had timestamps.")
+            # Note that only months (0 to 11) are used in OlmoEarth position encoding.
+            kwargs["timestamps"] = torch.stack(
+                [
+                    self.time_ranges_to_timestamps(time_range, max_timesteps, device)
+                    for time_range in timestamps_per_instance
+                ],
+                dim=0,
```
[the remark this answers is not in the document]
gabrieltseng: I added a note about this [here](https://github.com/allenai/rslearn/blob/6594140e424946d891336bd0e8e129422eeb2315/rslearn/models/olmoearth_pretrain/model.py#L227) - is this enough or should i add a more emphatic warning / todo?

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -61,6 +62,7 @@ def __init__(
         embedding_size: int | None = None,
         autocast_dtype: str | None = "bfloat16",
         token_pooling: bool = True,
+        use_legacy_timestamps: bool = False,
```
favyen2: I'm still a bit worried about the impact on existing models. What do you think about having this default to True but it shows a deprecation warning if the user has not passed use_legacy_timestamps=False? This way we can continue to use existing models with their existing model configurations, while new models will be encouraged to set this flag since otherwise it will show warning (similar warning to https://github.com/allenai/rslearn/blob/b4bbae662540f5ce7c01d3ac601a0ea60cba1c6b/rslearn/config/dataset.py#L237). Or if you test existing model and it doesn't really care about the timestamp (same performance with/without this flag) then I think it's fine; would be interesting to test new models too to see if the timestamp matters at all.

rslearn/models/presto/presto.py
```diff
@@ -184,6 +178,23 @@ def construct_presto_input(
             x = (x + PRESTO_ADD_BY.to(device=device)) / PRESTO_DIV_BY.to(device=device)
         return x, mask, dynamic_world.long(), months.long()
 
+    @staticmethod
+    def time_ranges_to_timestamps(
+        time_ranges: list[tuple[datetime, datetime]],
+        device: torch.device,
+    ) -> torch.Tensor:
+        """Turn the time ranges stored in a RasterImage to timestamps accepted by Galileo.
```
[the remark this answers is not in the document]
gabrieltseng: updated

rslearn/train/transforms/crop.py
```diff
@@ -80,13 +84,23 @@ def apply_image(self, image: torch.Tensor, state: dict[str, Any]) -> torch.Tenso
...
         remove_from_left = state["remove_from_left"] * image.shape[-1] // image_shape[1]
         remove_from_top = state["remove_from_top"] * image.shape[-2] // image_shape[0]
-        return torchvision.transforms.functional.crop(
-            image,
-            top=remove_from_top,
-            left=remove_from_left,
-            height=crop_size,
-            width=crop_size,
-        )
+        if isinstance(image, RasterImage):
```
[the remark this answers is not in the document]
gabrieltseng: I guess its a bit confusing because RasterImage has an additional `T` dimension which typically won't make sense for the targets (and then it introduces all sorts of weird indexing, since we'd constantly need to be taking the 0th timestep of the targets, which is probably more of a nuisance than the code here)

rslearn/train/transforms/concatenate.py
```diff
@@ -14,17 +17,20 @@ def __init__(
...
         output_selector: str,
+        concatenate_dim: int = 0,
     ):
         """Initialize a new Concatenate.
 
         Args:
             selections: map from selector to list of band indices in that input to
                 retain, or empty list to use all bands.
             output_selector: the output selector under which to save the concatenate image.
+            concatenate_dim: the dimension against which to concatenate the inputs
```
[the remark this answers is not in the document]
gabrieltseng: updated

rslearn/train/model_context.py
```diff
@@ -10,6 +10,39 @@
 from rslearn.utils.geometry import PixelBounds, Projection
 
 
+@dataclass
+class RasterImage:
+    """A raster image is a torch.tensor containing the images and their associated timestamps."""
+
+    # image will have a temporal dimension, CTHW
```
favyen2: I think something like "# image is a 4D CTHW tensor." is more clear, I don't understand this comment and why it says the temporal dimension in particular.

favyen2 (approved):

rslearn/train/model_context.py
gabrieltseng: updated

rslearn/train/all_patches_dataset.py
```diff
@@ -294,12 +294,24 @@ def crop_input_dict(d: dict[str, Any]) -> dict[str, Any]:
                                     int(end_offset[0] * scale),
                                     int(end_offset[1] * scale),
                                 )
-                                # Crop the CHW tensor with scaled coordinates.
-                                cropped[input_name] = value[
-                                    :,
-                                    scaled_start[1] : scaled_end[1],
-                                    scaled_start[0] : scaled_end[0],
-                                ].clone()
+                                if isinstance(value, torch.Tensor):
```
favyen2: Since this logic is repeated in two places, suggest to move the cropping of a `torch.Tensor | RasterImage` to a separate helper function.

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -86,7 +88,13 @@ def __init__(
             token_pooling: whether or not to pool the tokens. If True, the output will be BxCxHxW. If False,
                 there will be an extra dimension, N, (BxCxHxWxN) representing the temporal and channel
                 dimensions.
+            use_legacy_timestamps: In our original implementation of OlmoEarth, we applied timestamps starting
+                from 0 (instead of the actual timestamps of the input). The option to do this is preserved
+                for backwards compatability with finetuned models which were trained against this implementation.
         """
+        if use_legacy_timestamps:
+            raise DeprecationWarning("For new projects, don't use legacy timesteps.")
```
favyen2: Does this work? It looks like it would just error out if use_legacy_timestamps is set to True but I was proposing to show warning message instead. But maybe it works idk.

rslearn/train/transforms/resize.py
```diff
@@ -48,15 +52,24 @@ def apply_resize(self, image: torch.Tensor) -> torch.Tensor:
...
             image = image.unsqueeze(0)  # (H, W) -> (1, H, W)
             result = torchvision.transforms.functional.resize(
                 image, self.target_size, self.interpolation
             )
             return result.squeeze(0)  # (1, H, W) -> (H, W)
 
-        return torchvision.transforms.functional.resize(
-            image, self.target_size, self.interpolation
-        )
+        if isinstance(image, torch.Tensor):
```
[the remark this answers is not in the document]
gabrieltseng: updated

rslearn/train/all_patches_dataset.py
gabrieltseng: updaed

rslearn/train/dataset.py
```diff
@@ -284,7 +293,36 @@ def read_raster_layer_for_data_input(
             src[src_indexes, :, :].astype(data_input.dtype.get_numpy_dtype())
         )
 
-    return image
+    # add the timestamp. it can be the middle of the time range but for now
+    # lets just make it the beginning. TODO is to update that
+    time_range = None
+    if layer_data is not None:
+        if "time_range" in layer_data.serialized_item_groups[group_idx][0].get(
```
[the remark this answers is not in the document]
gabrieltseng: updated

PR #399, opened 2025-12-18
Add data source for the AWS s3://sentinel-cogs bucket maintained by Element 84.

favyen2:
The data source is very similar to Planetary Computer for get_items implementation, so I created an abstract StacSource to handle cases where basically the get_items is just served from a STAC API. I also implemented a StacClient to replace our usage of pystac_client in cases where we are just calling the search endpoint only; the client only supports that one endpoint, but doesn't have the caching issues since the only thing it does beyond making a request is handle pagination in the response.

Also it seems the `aws_open_data.Sentinel2` bucket is now free so I updated it to use anonymous access for that. That one has a STAC API so in theory we could greatly simplify get_items implementation by calling that API but I left it otherwise as is for now.

### DIFF
--- a/rslearn/data_sources/aws_sentinel2_element84.py
+++ b/rslearn/data_sources/aws_sentinel2_element84.py
@@ -52,7 +52,7 @@ class Sentinel2(StacDataSource, TileStore):
         "nir": ["B08"],
         "nir09": ["B09"],
         "swir16": ["B11"],
-        "swin22": ["B12"],
+        "swir22": ["B12"],
         "nir08": ["B8A"],
         "visual": ["R", "G", "B"],
     }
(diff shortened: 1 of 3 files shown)


rslearn/data_sources/aws_sentinel2_element84.py
```diff
@@ -0,0 +1,318 @@
...
+        Args:
+            layer_name: the layer name or alias.
+            item_name: the item.
+            bands: the list of bands identifying which specific raster to read.
+
+        Returns:
+            whether there is a raster in the store matching the source, item, and
+                bands.
+        """
+        # Always ready since we wrap accesses to Planetary Computer.
```
yawenzzzz: remove "Planetary Computer"

yawenzzzz (approved):
LGTM! just a few minor things to fix.

PR #400, opened 2025-12-18
Allow dataset config to be passed for dataset init

cmwilhelm:
* used with precedence over `path / config.json`
* bit of validation logic moved out into pydantic model

### DIFF
--- a/rslearn/dataset/dataset.py
+++ b/rslearn/dataset/dataset.py
@@ -43,29 +43,26 @@ class Dataset:
     def __init__(
         self,
         path: UPath,
-        dataset_config: DatasetConfig | None = None,
         disabled_layers: list[str] = [],
+        dataset_config: DatasetConfig | None = None,
     ) -> None:
         """Initializes a new Dataset.
 
         Args:
             path: the root directory of the dataset
-            dataset_config: optional dataset configuration to use instead of loading from the dataset directory
             disabled_layers: list of layers to disable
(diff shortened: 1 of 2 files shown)


rslearn/dataset/dataset.py
```diff
@@ -40,37 +40,46 @@ class Dataset:
...
-                    self.path
-                )
+        config_content: str
+        if dataset_config is None:
+            # Load dataset configuration from the dataset directory.
+            with (self.path / "config.json").open("r") as f:
+                config_content = f.read()
+        else:
+            # Reserialize so we can perform template substitution on the config.
+            config_content = json.dumps(dataset_config.model_dump(mode="json"))
```
favyen2: Hm would it make sense for env substitution to be the responsibility of the caller if they are passing the DatasetConfig directly? I feel like it would make sense to assume it has already been done here. I'm okay with it either way though, but if it is substituted in the init function then that should probably be documented in the docstring.

cmwilhelm: Hm yeah, maybe that's more reasonable to make it caller's responsibility. I went back and forth on this but I think you've convinced me the other way.

favyen2: I was also thinking just now in theory you could implement arguably "better" environment var substitution closer to the transpiler since you can do the substitution either during transpiling or at least on the DatasetConfig struct (so can substitute entire str variables instead of within the whole config string). Not sure if that is really better or not but anyway opens up more options I guess.

PR #413, opened 2026-01-14
Hadriens/seg metrics per class

Farbum:
This PR adds:

- Dice loss for segmentation task
- X-entropy optional weights for class imbalance
- optional reporting of metric per class for segmentation metrics

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__
 .vscode
 build
 rslearn.egg-info
+.DS_Store
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ for how to setup these data sources.
 - Xyz (Slippy) Tiles (e.g., Mapbox tiles)
 - Planet Labs (PlanetScope, SkySat)
 - ESA WorldCover 2021
+- ISRIC SoilGrids (WCS)
 
 rslearn can also be used to easily mosaic, crop, and re-project any sets of local
 raster and vector files you may have.
--- a/docs/DatasetConfig.md
+++ b/docs/DatasetConfig.md
@@ -1172,6 +1172,34 @@ Available bands:
 - B10
 - B11
 
+### rslearn.data_sources.soilgrids.SoilGrids
+
+This data source provides access to [ISRIC SoilGrids](https://www.isric.org/explore/soilgrids)
+via the public WCS endpoints (e.g. `https://maps.isric.org/mapserv?map=/map/clay.map`).
+
+This source is intended for **direct materialization** (set `"ingest": false` in the
+layer's `data_source` config), since data is fetched on-demand per window.
+
+Example (clay, user-provided WCS subset parameters):
+
+```jsonc
+{
+  "service_id": "clay",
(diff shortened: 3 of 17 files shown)


rslearn/train/tasks/segmentation.py
```diff
@@ -332,7 +406,7 @@ def __init__(
         self,
         metric: Metric,
         pass_probabilities: bool = True,
-        class_idx: int | None = None,
```
favyen2: It seems like this will break existing configs such as https://github.com/allenai/olmoearth_projects/blob/b9329999162a6265566cf0f3e1fbc66454194577/olmoearth_run_data/nandi/model.yaml#L67

Is it possible to deprecate the class_idx setting instead of removing it entirely? Or keep it because it could be useful if users want to apply various torchmetrics which do return a tensor instead of dict (so should be indexed by integer, not string).

rslearn/train/tasks/segmentation.py
```diff
@@ -584,9 +658,11 @@ def compute(self) -> Any:
         """Compute metric.
 
         Returns:
-            the mean IoU across classes.
+            dict with "avg" containing the mean IoU across classes, and
```
favyen2: Do you think it'd be worthwhile to follow the convention in torchmetrics like https://lightning.ai/docs/torchmetrics/stable/classification/precision.html#torchmetrics.classification.MulticlassPrecision and have argument to the constructor to control the averaging? Then it can either return a scalar mean IoU here or a 1D tensor of per-class mean IoUs. And the SegmentationMetric can work with both built-in metrics and the custom ones we've defined here.

rslearn/train/data_module.py
```diff
@@ -231,6 +231,7 @@ def _get_dataloader(
                 num_replicas=self.trainer.world_size,
                 rank=self.trainer.global_rank,
                 shuffle=should_shuffle,
+                drop_last=(split == "val"),  # Ensure even batches for validation
```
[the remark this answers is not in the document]
Farbum: I was about to remove this line. It fixes a bug that occurs with multi-gpu fine-tuning.
This is part of a bigger problem that I am exploring in a different PR, so I am taking it off this PR.

rslearn/train/tasks/segmentation.py
Farbum: Good point, added the logic to keep both

rslearn/train/tasks/segmentation.py
Farbum: Ok I read your comment too fast.
So unfortunately lightning does not handle nested dictionaries (a metric collection item returning a dictionary). 
This is why I opted first to create one metric object per metric per class.

That being said, there is a way to do it that requires minor changes to lightning and multi_task.
- lightning has to first compute the metrics via the collection (which flattens the nest dict), then we log. That requires to move the logging from validation_step to on_validation_epoch_end and manually compute/reset the metrics.
- multi_task needs to flatten the structure so there's only one MetricCollection with all metrics at the top level.

I tested it and it works. Not too sure which option is cleanest, though.
What do you think?

Issue #423, opened 2026-01-16
PC datasource to support cop-dem


robmarkcole:
Since SRTM is [flaky](https://github.com/allenai/rslearn/issues/416), support cop-dem via PC - https://planetarycomputer.microsoft.com/dataset/cop-dem-glo-30

PR #424, opened 2026-01-16
Add CopDemGlo30 data source and associated tests

robmarkcole:
Address https://github.com/allenai/rslearn/issues/423

- Implement CopDemGlo30 class for accessing Copernicus DEM GLO-30 data.
- Enhance StacDataSource with time range handling for static datasets.
- Add unit tests for CopDemGlo30 to validate context layer configuration and error handling.
- Updated documentation, including for the undocumented NAIP source

Example config.json
```json
{
  "layers": {
    "dem": {
      "type": "raster",
      "band_sets": [
        {
          "dtype": "float32",
          "bands": [
            "DEM"
          ]
        }
      ],
      "data_source": {
        "class_path": "rslearn.data_sources.planetary_computer.CopDemGlo30",
        "init_args": {}
      }
    }
  }
}
```

### DIFF
--- a/docs/DatasetConfig.md
+++ b/docs/DatasetConfig.md
@@ -600,16 +600,21 @@ Available bands:
 - vv
 - vh
 
-### rslearn.data_sources.climate_data_store.ERA5LandMonthlyMeans
+### rslearn.data_sources.climate_data_store.ERA5Land
 
-This data source is for ingesting ERA5 land monthly averaged data from the Copernicus Climate Data Store.
+Base class for ingesting ERA5 land data from the Copernicus Climate Data Store.
 
-We recommend using the default number of workers (`--workers 0`, which means using the
-main process only) and batch size equal to the number of windows when preparing the
-ERA5LandMonthlyMeans dataset, as it will combine multiple geometries into a single CDS
-API request for each month to speed up dataset ingestion.
+An API key must be provided either in the data source configuration or via the CDSAPI_KEY
+environment variable. You can acquire an API key by registering at the Climate Data Store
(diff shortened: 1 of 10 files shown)


docs/DatasetConfig.md
```diff
@@ -1120,6 +1120,48 @@ Available bands:
...
 Note that B10 is not present in L2A.
 
+### rslearn.data_sources.planetary_computer.Naip
+
+NAIP imagery on Microsoft Planetary Computer. Direct materialization is supported.
+
+This data source uses the Planetary Computer `naip` collection, and reads the `image`
+asset which contains four bands: `R`, `G`, `B`, `NIR`.
+
+Note: this class currently always ingests/reads the full 4-band asset as a single band
```
favyen2: The asset on Planetary Computer is 4-band so I think even in the base class we don't provide a way to configure it with a different band grouping if the user is ingesting the assets.

For materialization, I think both the base class and the Naip subclass would work to read a subset if the user configures the band sets that way (so the asset bands specifies R,G,B,NIR, but in the band_sets key in dataset config they could just get NIR for example).

robmarkcole:
@favyen2 not sure about these failures
```
#19 ERROR: denied: installation not allowed to Write organization package
------
 > pushing ghcr.io/allenai/rslearn:sha-9a50a9945e58c2c492d311c7583f7e23b4dabda9 with docker:
------
ERROR: failed to build: denied: installation not allowed to Write organization package
Reference
Check build summary support
Error: buildx failed with: ERROR: failed to build: denied: installation not allowed to Write organization package
```

PR #429, opened 2026-01-16
Henryh/allow context set via env var

Hgherzog:
This pull request updates the multiprocessing context selection logic in `rslearn/main.py` to allow configuration via an environment variable, making it easier to customize process spawning behavior without code changes.

Multiprocessing configuration improvements:

* Added a new constant `MP_CONTEXT_ENV_VAR` to define the environment variable name for customizing the multiprocessing context.
* Changed the initialization of the multiprocessing context in the `main()` function to read from the `RSLEARN_MULTIPROCESSING_CONTEXT` environment variable if set, otherwise defaulting to `"forkserver"`.
* Imported the `os` module to support reading environment variables for multiprocessing context configuration.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ __pycache__
 .vscode
 build
 rslearn.egg-info
+.DS_Store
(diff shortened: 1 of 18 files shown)


favyen2 (approved):
Approving this but spawn is not recommended, can you share the dataset and model config where you need to set it to spawn? There should be a way to fix it.

PR #436, opened 2026-01-21
Visualization module

piperwolters:
A first attempt at a generic visualization module for `rslearn_projects` datasets. 

This starts a web server that displays an html visualization of the specified dataset. If you run this locally, you'll go to localhost:{port}. If you run this on a remote server, you'll need to have the `--host-networking` flag and then you'll go to {server}:{port}.

The visualization itself will show you specified image layers as well as the following for different task types:
- classification : text label 
- segmentation : per pixel mask, where no_data=black
- object detection : bounding boxes for each object

Limitations:
- There is currently just normalization implemented for sentinel2
- Regression is untested, but should work the same as classification and just show a text label.

NOTE: your dataset config **must** contain `class_property_name` and `class_names`

Examples:
- classification
```
python -m rslearn.vis.server $DATASET_PATH \
    --layers sentinel2 \
    --label_layers label  \
    --bands '{"sentinel2": ["B04", "B03", "B02"]}' \
    --normalization '{"sentinel2": "sentinel2_rgb"}' \
    --task_type classification \
    --max_samples 100
```
 
- segmentation
```
python -m rslearn.vis.server $DATASET_PATH \
    --layers sentinel2 \
    --label_layers label_raster \
    --bands '{"sentinel2": ["B04", "B03", "B02"]}' \
    --normalization '{"sentinel2": "sentinel2_rgb"}' \
    --task_type segmentation \
    --max_samples 10 --port 8000
```

- object detection
```
python -m rslearn.vis.server $DATASET_PATH \
  --layers sentinel2 \
  --label_layers label \
  --bands '{"sentinel2": ["B04", "B03", "B02"]}' \
  --normalization '{"sentinel2": "sentinel2_rgb"}' \
  --task_type detection \
  --max_samples 100
```

### DIFF
--- a/rslearn/vis/render_raster_label.py
+++ b/rslearn/vis/render_raster_label.py
@@ -1,7 +1,5 @@
 """Functions for rendering raster label masks (e.g., segmentation masks)."""
 
-from io import BytesIO
-
 import numpy as np
 from PIL import Image
 from rasterio.warp import Resampling
@@ -12,17 +10,15 @@ from rslearn.log_utils import get_logger
 from rslearn.train.dataset import DataInput, read_raster_layer_for_data_input
 from rslearn.utils.geometry import PixelBounds, ResolutionFactor
 
-from .utils import VISUALIZATION_IMAGE_SIZE
-
 logger = get_logger(__name__)
 
 
-def render_raster_label_as_bytes(
+def render_raster_label(
     label_array: np.ndarray,
     label_colors: dict[str, tuple[int, int, int]],
     layer_config: LayerConfig,
-) -> bytes:
-    """Render a raster label array as a colored mask PNG bytes.
+) -> np.ndarray:
+    """Render a raster label array as a colored mask numpy array.
 
     Args:
         label_array: Raster label array with shape (bands, height, width) - typically single band
@@ -30,7 +26,7 @@ def render_raster_label_as_bytes(
         layer_config: LayerConfig object (to access class_names if available)
 
     Returns:
-        PNG image bytes
+        Array with shape (height, width, 3) as uint8
     """
     if label_array.ndim == 3:
         label_values = label_array[0, :, :]
@@ -54,11 +50,7 @@ def render_raster_label_as_bytes(
         mask_img[mask] = color
 
     img = Image.fromarray(mask_img, mode="RGB")
-    img = img.resize(VISUALIZATION_IMAGE_SIZE, Image.Resampling.NEAREST)
-
-    buf = BytesIO()
-    img.save(buf, format="PNG")
-    return buf.getvalue()
+    return np.array(img)
 
 
 def read_raster_layer(
--- a/rslearn/vis/render_sensor_image.py
+++ b/rslearn/vis/render_sensor_image.py
@@ -1,38 +1,27 @@
 """Functions for rendering raster sensor images (e.g., Sentinel-2, Landsat)."""
 
-from io import BytesIO
-
 import numpy as np
-from PIL import Image
 
 from .normalization import normalize_array
-from .utils import VISUALIZATION_IMAGE_SIZE
 
 
-def render_sensor_image_as_bytes(
+def render_sensor_image(
     array: np.ndarray,
     normalization_method: str,
-) -> bytes:
-    """Render a raster sensor image array as PNG bytes.
+) -> np.ndarray:
+    """Render a raster sensor image array as a numpy array.
 
     Args:
         array: Array with shape (channels, height, width) from RasterFormat.decode_raster
         normalization_method: Normalization method to apply
 
     Returns:
-        PNG image bytes
+        Array with shape (height, width, channels) as uint8
     """
     normalized = normalize_array(array, normalization_method)
 
-    if normalized.shape[-1] == 1:
-        img = Image.fromarray(normalized[:, :, 0], mode="L")
-    elif normalized.shape[-1] == 3:
-        img = Image.fromarray(normalized, mode="RGB")
-    else:
-        img = Image.fromarray(normalized[:, :, :3], mode="RGB")
+    # If more than 3 channels, take only the first 3 for RGB
+    if normalized.shape[-1] > 3:
+        normalized = normalized[:, :, :3]
 
-    img = img.resize(VISUALIZATION_IMAGE_SIZE, Image.Resampling.LANCZOS)
-
-    buf = BytesIO()
-    img.save(buf, format="PNG")
-    return buf.getvalue()
+    return normalized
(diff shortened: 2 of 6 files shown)


piperwolters:
Sample htmls (saved by adding the `--save_html` flag):
[sample_geo_annual.html](https://github.com/user-attachments/files/24754460/sample_geo_annual.html)
[sample_worldcover.html](https://github.com/user-attachments/files/24754461/sample_worldcover.html)
[sample_kenya_maize_cropland_segmentation.html](https://github.com/user-attachments/files/24754462/sample_kenya_maize_cropland_segmentation.html)
[sample_kenya_maize_crop_classification.html](https://github.com/user-attachments/files/24754464/sample_kenya_maize_crop_classification.html)
[sample_wind_turbines.html](https://github.com/user-attachments/files/24754492/sample_wind_turbines.html)

rslearn/vis/labels.py
```diff
@@ -0,0 +1,294 @@
...
+    Returns:
+        Dictionary mapping label class names to RGB color tuples
+    """
+    label_classes = sorted(label_classes)
+    label_colors = {}
+
+    # Hardcoded color for "no_data" (always black - rslearn convention)
+    NO_DATA_COLOR = (0, 0, 0)
+
+    colors = [
```
favyen2: We could move https://github.com/allenai/rslearn/blob/23f873df2105d57f378d7818ad5f6487fa6d5183/rslearn/train/tasks/segmentation.py#L24 to somewhere in `rslearn.utils` (maybe a new module like `rslearn/utils/colors.py`) and then reuse that there, in `detection.py`, and here.

rslearn/vis/normalization.py
```diff
@@ -0,0 +1,66 @@
+"""Normalization functions for raster data visualization."""
+
+import numpy as np
+
+from rslearn.log_utils import get_logger
+
+logger = get_logger(__name__)
+
+
+def normalize_band(band: np.ndarray, method: str = "sentinel2_rgb") -> np.ndarray:
```
favyen2: Suggest to make method an Enum and move the per-method code to different functions. Then this function could make calls to those functions, either via the if statements like you have or a dict that maps from NormalizationMethod to `Callable[[np.ndarray], np.ndarray]`.

rslearn/vis/normalization.py
```diff
@@ -0,0 +1,66 @@
...
+    """Normalize a multi-band array to 0-255 range.
+
+    Args:
+        array: Input array with shape (bands, height, width) or (height, width, bands)
+        method: Normalization method (applied per-band)
+
+    Returns:
+        Normalized array as uint8
+    """
+    # Handle (bands, height, width) format
```
favyen2: `RasterFormat.decode_raster` always returns a CHW array so there should not be a need to auto-detect the array shape here.

rslearn/vis/vis.py
```diff
@@ -0,0 +1,400 @@
...
+from rasterio.warp import transform
+
+from rslearn.log_utils import get_logger
+from rslearn.utils.feature import Feature
+from rslearn.utils.geometry import PixelBounds, Projection
+
+from .normalization import normalize_array
+
+if TYPE_CHECKING:
+    from rslearn.config import LayerConfig
```
favyen2: I don't think there is any risk of import loop here so it shouldn't need the `TYPE_CHECKING` flag (can just always import).

rslearn/vis/server.py
```diff
@@ -0,0 +1,794 @@
+"""
```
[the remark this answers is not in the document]
piperwolters: Took a stab at implementing this.

rslearn/vis/server.py
```diff
@@ -0,0 +1,1128 @@
...
+                            bands[ref_layer_name],
+                            group_idx=group_idx,
+                        )
+                        ref_normalization_method = normalization[ref_layer_name]
+
+                width = window.bounds[2] - window.bounds[0]
+                height = window.bounds[3] - window.bounds[1]
+
+                if reference_array is not None:
+                    from PIL import ImageDraw
```
favyen2: Move imports to top of file.

rslearn/vis/server.py
```diff
@@ -0,0 +1,1128 @@
...
+
+class VisualizationServer:
+    """Visualization server for rslearn datasets using Dataset/Window APIs."""
+
+    def __init__(self) -> None:
+        """Initialize the visualization server."""
+        self.windows: list[Window] = []
+        self.dataset: Dataset | None = None
+        self.layers: list[str] = []
+        self.bands: dict[str, list[str]] = {}
```
favyen2: Add comment saying what the keys and values are for these dicts.

rslearn/vis/server.py
```diff
@@ -0,0 +1,1128 @@
...
+                                ):
+                                    mask_layers.append(layer_name)
+                        elif layer_config.type == LayerType.RASTER:
+                            # For raster label layers, we add them if completed
+                            # The image endpoint will handle empty data by returning 404
+                            if window.is_layer_completed(
+                                layer_name, group_idx=self.group_idx
+                            ):
+                                mask_layers.append(layer_name)
+                except Exception as e:
```
favyen2: What errors happen here? I just wonder if this is needed, and if so whether the errors can be caught earlier so they don't need to be caught here.

rslearn/vis/server.py
```diff
@@ -0,0 +1,1128 @@
...
+                                    )
+
+                                if server_instance.dataset is None:
+                                    self.send_response(500)
+                                    self.send_header("Content-type", "text/plain")
+                                    self.end_headers()
+                                    self.wfile.write(b"Server not initialized")
+                                    return
+
+                                image_bytes = server_instance._generate_image_as_bytes(
```
favyen2: I think it might be somewhat more readable to have the generate_image function return an image and then convert it to bytes in the HTTP handler.

rslearn/vis/server.py
```diff
@@ -0,0 +1,1128 @@
...
+            band_names = [b.strip() for b in part.split(",")]
+            if bands_dict:
+                raise ValueError(
+                    "Cannot use single-layer format with multiple layers. Use layer:band1,band2 format"
+                )
+            bands_dict["_default"] = band_names
+    return bands_dict
+
+
+def parse_normalization_arg(norm_str: str | None) -> dict[str, str]:
```
favyen2: Maybe these can just take JSON instead of having complex parsing.

favyen2 (commented):
Thanks for the updates! The normalization code looks a lot better to me and I like it returning images from a separate handler rather than embedding them into the HTML.

I had several comments but they are minor, my main concerns are with the `server.py` I think can be improved by using flask and breaking down large functions with lots of cases into multiple functions, and in `vis.py` / `overlay.py` to reorganize the code so there is one module per logical group of functions like `render_sensor_image.py`, `render_vector_label.py`, `render_raster_label.py` -- I saw you moved some code to `overlay.py` but I was thinking more to include the vector code together but in different place from handling raster label code which can be different place from handling raster image code.

PR #442, opened 2026-01-22
Hadriens/writing_test_metrics

Farbum:
Small PR to write metrics to the parent folder of the checkpoint path by default


Farbum:
> @Farbum is the idea to change this PR to keep the automatic test metric location but remove the group thing?

Yes, I did but I think there are some relations to the other logging bug we have talked about. I think I need to take some time to dive into the logging process etc.
Leaving it as a draft PR for now, will come back to it when I am done with the WF stuff

PR #444, opened 2026-01-22
Fix data source tests

favyen2:
The weekly data source integration tests have been failing for a while. This PR tries to fix them.

Some of the errors it has run into are legitimate -- for example, the URL used by `earthdata_srtm.py` has been decommissioned (https://e4ftl01.cr.usgs.gov/MEASURES/SRTMGL1.003/). So I do think these tests are worth keeping around. For SRTM I have created a mirror of the SRTM data on Hugging Face.

### DIFF
--- a/.github/workflows/data-source-integration-test.yml
+++ b/.github/workflows/data-source-integration-test.yml
@@ -75,17 +75,13 @@ jobs:
             username: ${{ github.actor }}
             password: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Authenticate into gcp
-        uses: "google-github-actions/auth@v2"
-        with:
-          credentials_json: ${{secrets.GOOGLE_CREDENTIALS }}
-
       - name: Run Data Sources Integration Tests
         run: |
+          printenv GOOGLE_CREDENTIALS > /tmp/gcp-credentials.json
           docker run --rm \
             -e TEST_BUCKET="earthsystem-dev-c3po-rslearn-test-bucket" \
             -e TEST_PREFIX="rslearn_gha_data_source_tests/" \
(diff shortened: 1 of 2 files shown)


rslearn/utils/retry_session.py
```diff
@@ -0,0 +1,43 @@
...
+        allowed_methods=["GET", "POST"],
+        status_forcelist=[
+            HTTPStatus.TOO_MANY_REQUESTS,
+            HTTPStatus.INTERNAL_SERVER_ERROR,
+            HTTPStatus.BAD_GATEWAY,
+            HTTPStatus.SERVICE_UNAVAILABLE,
+            HTTPStatus.GATEWAY_TIMEOUT,
+        ],
+        raise_on_status=False,
+    )
```
APatrickJ: Probably good to add `respect_retry_after_header=True` here

rslearn/utils/m2m_api.py
```diff
@@ -0,0 +1,264 @@
...
+        while True:
+            time.sleep(10)
+            response_dict = self.request("download-retrieve", {"label": label})
+            if response_dict is None:
+                raise APIException("No response from API")
+            response = response_dict["data"]
+            if len(response["available"]) > 0:
+                return response["available"][0]["url"]
+            if len(response["requested"]) == 0:
+                print(response)
```
APatrickJ: Do you want to keep this print?

rslearn/utils/retry_session.py
favyen2: It looks like that is the default, do you think it is worth adding it explicitly? There are some other arguments like `raise_on_redirect=True` so I think it could make sense to leave out the ones where we are using the default.

PR #458, opened 2026-01-26
Tilestore API code simplification

favyen2:
Previously we have some redundant code across data sources that support direct materialization, to implement the parts of TileStore that are needed. Now this redundant code is moved to a `TileStoreDataSource` base class, and most data sources supporting direct materialization can inherit from `TileStoreDataSource` and use its functionality while just implementing a couple of functions (e.g. to get the URL to download an asset from).

Other data sources like `XyzTiles` still directly inherit `TileStore` since they have pretty different architectures.


rslearn/data_sources/tile_store_data_source.py
```diff
@@ -0,0 +1,336 @@
...
+
+from rslearn.config import LayerConfig
+from rslearn.data_sources.data_source import DataSource, ItemType
+from rslearn.dataset import Window
+from rslearn.dataset.materialize import RasterMaterializer
+from rslearn.tile_stores import TileStore, TileStoreWithLayer
+from rslearn.utils.geometry import PixelBounds, Projection
+
+
+class TileStoreDataSource(DataSource[ItemType], TileStore, Generic[ItemType]):
```
APatrickJ: Optional: I found the name a bit confusing, as it sounds like it's a data source that uses a tile store, not a data source that is also a tile store. Could maybe consider naming it `DirectMaterializeDataSource`.

APatrickJ (approved):
Seems like a clean refactor to me! I made an optional comment about naming.

rslearn/data_sources/tile_store_data_source.py
favyen2: Thanks! Agreed that that is better, renamed.

PR #463, opened 2026-01-27
Add confusion matrix

yawenzzzz:
This PR adds confusion matrix logging to wandb for classification and segmentation tasks, with a new `NonScalarMetricOutput` abstraction for handling non-scalar metrics. (i'm confused that why i'm not getting the Custom Charts, turns out it only appears when the run is complete). Also, i decided removing the figure as the Custom Charts is doing sth very similar.


rslearn/train/lightning_module.py
```diff
@@ -210,15 +211,48 @@ def on_train_epoch_start(self) -> None:
...
             pass
 
+    def _log_non_scalar_metric(self, name: str, value: NonScalarMetricOutput) -> None:
+        """Log a non-scalar metric to wandb.
+
+        Args:
+            name: the metric name (e.g., "val_confusion_matrix")
+            value: the non-scalar metric output
+        """
+        # Skip during Lightning sanity check
```
favyen2: Elaborate why this causes problems for sanity check. I assume it is because other metrics are handled by Lightning but we are directly logging here so we need to manually handle it?

rslearn/train/metrics.py
```diff
@@ -0,0 +1,104 @@
...
+import wandb
+
+from rslearn.log_utils import get_logger
+
+logger = get_logger(__name__)
+
+
+@dataclass
+class NonScalarMetricOutput(ABC):
+    """Base class for non-scalar metric outputs that need special logging.
```
favyen2: Clarify that these metrics will only support logging to W&B.

PR #465, opened 2026-01-27
Add ERA5 time-series

### LINKED ISSUE #427: Add ERA5 hourly time-series data
As discussed in #411, the ERA5 hourly data is slow to ingest (even with the bounds), therefore, it makes sense to add this ERA5 hourly time-series data (it requests a single lat/lon at a time): https://cds.climate.copernicus.eu/datasets/reanalysis-era5-land-timeseries?tab=overview.

yawenzzzz:
This PR adds the ERA5 Land hourly time-series data (https://cds.climate.copernicus.eu/datasets/reanalysis-era5-land-timeseries?tab=overview). Resolve #427

### DIFF
--- a/tests/unit/data_sources/test_climate_data_store.py
+++ b/tests/unit/data_sources/test_climate_data_store.py
@@ -1,9 +1,12 @@
+from unittest.mock import MagicMock, patch
+
 from rslearn.data_sources.climate_data_store import ERA5LandHourlyTimeseries
 
 TEST_BANDS = ["2m-temperature", "total-precipitation"]
 
 
-def test_grid_snapping() -> None:
+@patch("rslearn.data_sources.climate_data_store.cdsapi.Client")
+def test_grid_snapping(mock_client: MagicMock) -> None:
     """Test that coordinates are correctly snapped to 0.1 degree grid."""
     data_source = ERA5LandHourlyTimeseries(band_names=TEST_BANDS)
 


rslearn/data_sources/climate_data_store.py
```diff
@@ -469,3 +470,331 @@ def ingest(
                 tile_store.write_raster_file(
                     item.name, self.band_names, UPath(local_tif_fname)
                 )
+
+
+class ERA5LandHourlyTimeseries(DataSource):
+    """A data source for ingesting ERA5-Land hourly time-series data for individual points.
+
+    This data source corresponds to the reanalysis-era5-land-timeseries dataset, which is
```
favyen2: I think the docstring should say that it will always materialize a 1x1 for each window, with data about the closest 0.1 degree grid to the window's center.

rslearn/data_sources/climate_data_store.py
```diff
@@ -469,3 +470,331 @@ def ingest(
...
+        to enable proper caching and avoid duplicate requests.
+
+        Args:
+            lon: the longitude in degrees
+            lat: the latitude in degrees
+
+        Returns:
+            A tuple of (snapped_lon, snapped_lat) rounded to 1 decimal place.
+        """
+        snapped_lon = round(round(lon / self.PIXEL_SIZE) * self.PIXEL_SIZE, 1)
```
favyen2: Is this the same as `round(lon, 1)`?

I think it would be nice to support changing the PIXEL_SIZE and still having it work but in this case I think the code is already not general since it depends on rounding the result to first decimal at the end so might as well just do `round(lon, 1)` right?

yawenzzzz:
@favyen2 This PR is ready for a second look at.

tests/fixtures/geometries/seattlefeb2020.py
```diff
@@ -0,0 +1,28 @@
...
+import pytest
+import shapely
+from rasterio import CRS
+
+from rslearn.const import WGS84_PROJECTION
+from rslearn.utils import Projection, STGeometry
+
+
+@pytest.fixture
+def seattlefeb2020() -> STGeometry:
```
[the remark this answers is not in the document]
yawenzzzz: This test has been removed.

PR #466, opened 2026-01-28
temperature for softmax

Farbum:
Small PR to add temperature scaling in the segmentation softmax

### DIFF
--- a/rslearn/train/tasks/segmentation.py
+++ b/rslearn/train/tasks/segmentation.py
@@ -128,9 +128,7 @@ class SegmentationTask(BasicTask):
             return {}, {}
 
         assert isinstance(raw_inputs["targets"], RasterImage)
-        assert raw_inputs["targets"].image.shape[0] == 1
-        assert raw_inputs["targets"].image.shape[1] == 1
-        labels = raw_inputs["targets"].image[0, 0, :, :].long()
+        labels = raw_inputs["targets"].get_hw_tensor().long()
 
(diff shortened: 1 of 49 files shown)


rslearn/train/tasks/segmentation.py
```diff
@@ -291,19 +291,26 @@ def get_metrics(self) -> MetricCollection:
 class SegmentationHead(Predictor):
     """Head for segmentation task."""
 
-    def __init__(self, weights: list[float] | None = None, dice_loss: bool = False):
+    def __init__(
+        self,
+        weights: list[float] | None = None,
+        dice_loss: bool = False,
+        temperature: float = 1.0,
```
favyen2: Add test to verify output with temperature is correctly computed (e.g. try with two temperatures and two input logits make sure higher temperature makes the resulting probabilities less extreme). Also might be worth saying in the docstring that this will only affect the output probabilities, not the loss. Actually is that correct, should it be passed to the loss too, or is it only to change the output visualization?

Farbum: Yes agreed, it'd be good to have some tests for the Segmentation head, I'll add one.
As for the temperature affecting the loss, it was flowing into the Dice loss but I changed that so that the loss remains unaffected.

favyen2 (approved):
thanks for the update! and got it now I missed the dice loss thing but yeah makes sense either way as long as docstring matches up

PR #467, opened 2026-01-28
Fix W&B run resume by 

Farbum:
Sets resume='must' when resuming from checkpoint


favyen2:
According to the docs it should default to auto which should resume the run if it exists, is this PR so that it will throw error in case the run ID doesn't exist, or did you have issue where the run does exist in W&B but did not resume for some reason?

favyen2:
From what I can tell this W&B run resumed normally https://wandb.ai/eai-ai2/20260111_wf_burnrisk_nbac

PR #468, opened 2026-01-28
Add ERA5LandDailyUTCv1 data source

robmarkcole:
- Implemented the ERA5LandDailyUTCv1 class for accessing and ingesting ERA5-Land daily data from EarthDataHub.
- Added methods for handling dataset retrieval, item generation, and GeoTIFF writing.
- Created integration tests for the new data source, validating ingestion from a local Zarr store and ensuring correct raster output.

Question 1: Kelvin or Celcius default for temperature? 
Question 2: Add dask as dependency? If not we can set `chunks=None` (no dask)

Foundation for addressing https://github.com/allenai/rslearn/issues/427 CC @yawenzzzz 

Example config.json
```json
{
  "layers": {
    "era5_land_daily": {
      "type": "raster",
      "band_sets": [
        {
          "dtype": "float32",
          "bands": ["t2m", "tp"]
        }
      ],
      "data_source": {
        "class_path": "rslearn.data_sources.earthdatahub.ERA5LandDailyUTCv1",
        "init_args": {
          "temperature_unit": "kelvin"
        },
        "query_config": {
          "space_mode": "MOSAIC",
          "time_mode": "WITHIN"
        }
      }
    },
 ...
  },
  "tile_store": {
    "class_path": "rslearn.tile_stores.DefaultTileStore",
    "init_args": {
      "path_suffix": "tiles"
    }
  }
}
```

### DIFF
--- a/rslearn/data_sources/earthdatahub.py
+++ b/rslearn/data_sources/earthdatahub.py
@@ -2,10 +2,10 @@
 
 from __future__ import annotations
 
+import math
 import os
 import tempfile
 from datetime import UTC, datetime, timedelta
-import math
 from typing import Any, Literal
 
 import numpy as np
@@ -191,7 +191,9 @@ class ERA5LandDailyUTCv1(DataSource[Item]):
         elif band_names is not None:
             self.band_names = band_names
         else:
-            raise ValueError("band_names must be set if layer_config is not in the context")
+            raise ValueError(
+                "band_names must be set if layer_config is not in the context"
+            )
 
         invalid_bands = [b for b in self.band_names if b not in self.ALLOWED_BANDS]
         if invalid_bands:
@@ -202,7 +204,7 @@ class ERA5LandDailyUTCv1(DataSource[Item]):
 
         self._ds: Any | None = None
 
-    def _get_dataset(self):
+    def _get_dataset(self) -> xr.Dataset:
         """Open (and memoize) the backing ERA5-Land Zarr dataset."""
         if self._ds is not None:
             return self._ds
--- a/tests/integration/data_sources/test_earthdatahub_era5_land_daily.py
+++ b/tests/integration/data_sources/test_earthdatahub_era5_land_daily.py
@@ -1,19 +1,21 @@
+from datetime import UTC, datetime
+from pathlib import Path
+
 import numpy as np
 import pytest
 import rasterio
 import shapely
-from datetime import UTC, datetime
 
 
-def test_era5land_dailyutc_v1_ingest_from_local_zarr(tmp_path) -> None:
+def test_era5land_dailyutc_v1_ingest_from_local_zarr(tmp_path: Path) -> None:
     import xarray as xr
     import zarr  # noqa: F401
+    from upath import UPath
 
     from rslearn.const import WGS84_PROJECTION
     from rslearn.data_sources.earthdatahub import ERA5LandDailyUTCv1
     from rslearn.tile_stores import DefaultTileStore, TileStoreWithLayer
     from rslearn.utils.geometry import STGeometry
-    from upath import UPath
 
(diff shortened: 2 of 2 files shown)


rslearn/data_sources/earthdatahub.py
```diff
@@ -0,0 +1,357 @@
...
+) -> list[tuple[float, float]]:
+    """Convert lon bounds to one or two non-wrapping ranges in [0, 360]."""
+    min_lon, _, max_lon, _ = bounds
+
+    # Dateline-crossing bounds in [-180, 180] would have min_lon > max_lon.
+    if min_lon > max_lon:
+        raise ValueError(
+            "ERA5LandDailyUTCv1 does not yet support longitude ranges that cross the dateline "
+            f"(got bounds min_lon={min_lon}, max_lon={max_lon})."
+        )
```
yawenzzzz: Given that `bounds` are the output of `_snap_bounds_outward()`, we're probably not likely to get `min_lon > max_lon` right?

rslearn/data_sources/earthdatahub.py
```diff
@@ -0,0 +1,357 @@
...
+    Authentication:
+        EarthDataHub uses token-based auth. Configure your netrc file so HTTP clients
+        can attach the token automatically. On Linux and MacOS the netrc path is
+        `~/.netrc`.
+    """
+
+    DEFAULT_ZARR_URL = (
+        "https://data.earthdatahub.destine.eu/era5/era5-land-daily-utc-v1.zarr"
+    )
+    ALLOWED_BANDS = {"t2m", "tp"}
```
yawenzzzz: Why are we only allowing these two bands? There are more available in the dataset (e.g., wind, radiation) that could also be useful: https://earthdatahub.destine.eu/collections/era5/datasets/era5-land-daily

rslearn/data_sources/earthdatahub.py
```diff
@@ -0,0 +1,357 @@
...
+        invalid_bands = [b for b in self.band_names if b not in self.ALLOWED_BANDS]
+        if invalid_bands:
+            raise ValueError(
+                f"unsupported ERA5LandDailyUTCv1 band(s): {invalid_bands}; "
+                f"supported: {sorted(self.ALLOWED_BANDS)}"
+            )
+
+        self._ds: Any | None = None
+
+    def _get_dataset(self):
```
yawenzzzz: It would be nice to add docstrings to all functions to improve readability.

rslearn/data_sources/earthdatahub.py
```diff
@@ -0,0 +1,357 @@
...
+
+            start, end = geometry.time_range
+            cur_day = _floor_to_utc_day(start)
+            cur_groups: list[list[Item]] = []
+            while cur_day < end:
+                next_day = cur_day + timedelta(days=1)
+                item_name = f"era5land_dailyutc_v1_{cur_day.year:04d}{cur_day.month:02d}{cur_day.day:02d}"
+                item_geom = STGeometry(WGS84_PROJECTION, item_shp, (cur_day, next_day))
+                cur_groups.append([Item(item_name, item_geom)])
+                cur_day = next_day
```
yawenzzzz: I think saving the daily item is nice because with this the materialization will work properly since there're only two bands per item. Previously, we have discussed about how to save ERA5 time-series and decided to stack the hourly data into monthly as one item, which will be faster if we're materializing the full-month windows (which is true for a lot of our tasks).

rslearn/data_sources/earthdatahub.py
robmarkcole: added

rslearn/data_sources/earthdatahub.py
robmarkcole: Supporting all now

PR #470, opened 2026-01-28
bump to 0.026

Farbum:


favyen2 (approved):
Sounds good to me and approving but if you can include some context in the PR description it could be useful (in the future), I assume this is to get one of the recent changes like temperature though.

PR #471, opened 2026-01-28
Fix bug where not all pixels are covered when using overlap_pixels.

favyen2:
The issue is because the start position for the second crop was set correctly. We had a test to verify things work when we only have the first+last crops, but those are both special cases. So a new test is now added that verifies it works with first+last crop plus one crop in the middle.

Depends on #456


docs/examples/OlmoEarthEmbeddings.md
```diff
@@ -229,10 +229,11 @@ trainer:
...
-            # middle 8x8 of that by removing 4 pixels of padding from each side.
-            padding: 4
+            # This removes the border from the overlap_pixels. With model patch size 4
+            # and input crop size 64, the model produces a 16x16 output. We have
+            # overlap_pixels=32 at input resolution, which is 8 pixels at output
+            # resolution, so we remove 4 pixels from each side.
+            overlap_pixels: 8
             # Set this equal to patch size, so the merger expects the output from the
             # task to be at 1/(downsample_factor) resolution relative to the window
             # resolution.
```
yawenzzzz: Given that we may crop the input or extract the center of feature map, instead of saying `this equal to patch size`, maybe better to say this is an ratio between input and output resolution.

PR #472, opened 2026-01-30
Add data source for olmoearth_datasets

favyen2:
Currently I only implemented Sentinel-2 L2A subclass. Current implementation will shuffle the data providers and iterate over them in order, if all of them fail then retries are still handled outside the data source.

Depends on #458


favyen2:
Ah ok we are going to put this in internal olmoearth_run repo since it is using internal API.

PR #473, opened 2026-01-30
For Sentinel-2 harmonization, just parse scene ID instead of parsing product XML

favyen2:
Depends on #458.

This will make it simpler to do harmonization for data sources that do not provide the XML.

@APatrickJ I think strictly speaking we don't actually need this but it does simplify things. Do you think we should continue parsing XML or is this change good?


favyen2:
decide to keep it the way it was

PR #474, opened 2026-01-30
Clarify space modes in DatasetConfig.md and input keys in OlmoEarth.md.

favyen2:
This improves docs in two places:
- DatasetConfig.md: clean up the discussion about space mode, and add an example. I also removed time mode here (and deprecated the argument) since it doesn't seem to be useful.
- OlmoEarth.md: clarify that the input keys must match with those expected by the pre-trained models, and where the key names and band names for the OlmoEarthNormalize come from.

### DIFF
--- a/docs/DataSources.md
+++ b/docs/DataSources.md
@@ -1046,6 +1046,58 @@ default those to the window pixel size.
 Available bands:
 - B1 (float32 recommended; scale/offset applied; set `nodata_vals` to `-32768`)
 
+### rslearn.data_sources.soildb.SoilDB
+
+This data source reads OpenLandMap-SoilDB rasters from the [OpenLandMap static STAC
+catalog](https://stac.openlandmap.org/)
+
+Each SoilDB collection links to a single STAC Item which contains multiple GeoTIFF/COG
+assets (e.g., different depth ranges, resolutions, and summary statistics). rslearn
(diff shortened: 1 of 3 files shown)


docs/DatasetConfig.md
```diff
@@ -340,83 +338,104 @@ The data source specification looks like this:
...
+- INTERSECTS: use items that intersect the window bounds. As with CONTAINS, the
+  resulting item groups will each consist of exactly one item.
+- MOSAIC: create mosaics, where each item group combines multiple items from the data
+  source as needed to cover the entire window. In this case, each item group may
+  include multiple items. This strategy initializes a buffer of `max_matches` empty
+  item groups. It then iterates over the items, adding each item to the first group
+  that the item provides additional coverage for (skipping groups that already cover
+  all the portions of the window that the new item covers). Finally, the non-empty
+  groups are returned.
+- PER_PERIOD_MOSAIC: create one mosaic per sub-period of the time range. When using
```
cmwilhelm: What is the behavior of space_mode=CONTAINS and period_duration not null? In unified config I think I assumed that was possible, and thought it would result in CONTAINS behavior within each period, but seeing PER_PERIOD_MOSAIC and MOSAIC are defined separately makes me think that probably doesn't work the way I thought.

APatrickJ (approved):

PR #477, opened 2026-02-02
Small fix for concatenate features

yawenzzzz:
There is a tentative fix for the NaN data, I will try to resolve it in the data source too because those locations (outside ERA5 land boundary) should not get data (PR #465)

### DIFF
--- a/rslearn/data_sources/__init__.py
+++ b/rslearn/data_sources/__init__.py
@@ -17,6 +17,7 @@ from .data_source import (
     ItemLookupDataSource,
     RetrieveItemDataSource,
 )
+from .direct_materialize_data_source import DirectMaterializeDataSource
 
 __all__ = (
     "DataSource",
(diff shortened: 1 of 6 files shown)


rslearn/models/concatenate_features.py
```diff
@@ -78,8 +78,10 @@ def forward(self, intermediates: Any, context: ModelContext) -> FeatureMaps:
                 "Expected input to be FeatureMaps with at least one feature map"
             )
 
+        # image is a 4D CTHW tensor
         add_data = torch.stack(
-            [input_data[self.key] for input_data in context.inputs], dim=0
+            [input_data[self.key].image.flatten(0, 1) for input_data in context.inputs],
```
favyen2: suggest to use einops here to make the reshaping more clear

(but fine to keep like this, I know there are many places in rslearn codebase where we are reshaping without einops, hopefully we can gradually change them over to be more clear by using the einops function)

favyen2 (approved):
looks good to me!

rslearn/models/concatenate_features.py
yawenzzzz: ok, that make sense, that also remove the need for the additional comment

PR #488, opened 2026-02-05
Fix method resolution order bug

APatrickJ:
### Problem

When using direct materialization with `PlanetaryComputer` or `aws_sentinel2_element84.Sentinel2` data sources, calls to `get_item_by_name` were raising `NotImplementedError`:

```
File ".../rslearn/data_sources/direct_materialize_data_source.py", line 92, in get_item_by_name
    raise NotImplementedError
```

### Root Cause

`DirectMaterializeDataSource` defined its own `get_item_by_name` method that raised `NotImplementedError`, which shadowed the working implementation from `StacDataSource` due to Python's Method Resolution Order (MRO).

### Fix

Updated `DirectMaterializeDataSource` to inherit from `ItemLookupDataSource` instead of `DataSource`, and removed its separate `get_item_by_name` definition. This allows the MRO to naturally find `StacDataSource.get_item_by_name` for classes that inherit from both.

This is cleaner than requiring subclasses to explicitly delegate because:
- No boilerplate needed in subclasses
- Semantically correct: `DirectMaterializeDataSource` properly declares it needs item lookup capability via `ItemLookupDataSource`
- The inheritance hierarchy naturally resolves the method lookup

### DIFF
--- a/rslearn/data_sources/direct_materialize_data_source.py
+++ b/rslearn/data_sources/direct_materialize_data_source.py
@@ -18,7 +18,9 @@ from rslearn.tile_stores import TileStore, TileStoreWithLayer
 from rslearn.utils.geometry import PixelBounds, Projection
 
 
-class DirectMaterializeDataSource(ItemLookupDataSource[ItemType], TileStore, Generic[ItemType]):
+class DirectMaterializeDataSource(
+    ItemLookupDataSource[ItemType], TileStore, Generic[ItemType]
+):
     """Base class for data sources that support direct materialization via TileStore.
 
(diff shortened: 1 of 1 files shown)


favyen2 (commented):
Thanks for this fix! I think the issue is more that DirectMaterializeDataSource defines get_item_by_name without implementing it. StacDataSource inherits from ItemLookupDataSource but DirectMaterializeDataSource does not and instead defines get_item_by_name separately. If DirectMaterializeDataSource is updated to inherit from ItemLookupDataSource then I think it would address the problem more cleanly (without requiring subclasses to explicitly delegate).

PR #490, opened 2026-02-05
[Bugfix] Enhance PerPixelRegressionTask to handle window padding

robmarkcole:
### Issue
When `crop_bounds` extends outside window_bounds (e.g., fixed `crop_size` but small windows / edge crops), raster decoding pads the out-of-window area with zeros. `PerPixelRegressionTask.process_inputs()` previously built `targets["valid"]` solely from `nodata_value` equality **inside** the label raster, so padded zeros were treated as valid labels (especially when `nodata_value=-1`), contaminating loss and metrics (MSE/R²/Huber/L1) with out-of-window pixels.

### Resolution
`PerPixelRegressionTask.process_inputs()` now always masks out pixels outside the window by intersecting `metadata.crop_bounds` with `metadata.window_bounds` and AND-ing that “in-window” mask with the existing `nodata_value` validity mask. This ensures padded regions are automatically invalid for both loss and metrics, even when their padded value is not equal to `nodata_value`

### Test
Added `test_process_inputs_masks_out_of_window_padding` to reproduce a 20×20 crop around a 10×10 window and assert only the intersection (100 pixels) is marked valid.

### DIFF
--- a/tests/unit/train/tasks/test_per_pixel_regression.py
+++ b/tests/unit/train/tasks/test_per_pixel_regression.py
@@ -14,16 +14,12 @@ from rslearn.train.tasks.per_pixel_regression import (
 
 def _make_metadata(
     empty_sample_metadata: SampleMetadata,
-    height: int,
-    width: int,
+    window_bounds: tuple[int, int, int, int],
+    crop_bounds: tuple[int, int, int, int],
     *,
-    window_bounds: tuple[int, int, int, int] | None = None,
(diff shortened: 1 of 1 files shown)


favyen2:
BTW just to confirm, this is only an issue if the label GeoTIFF has incorrect nodata value right? Like label GeoTIFF nodata is 0 or not set because the GeoTIFF was materialized using rslearn from a data source (we don't track nodata value when materializing), and so it ends up being 0 when reading outside the window bounds?

PR #494, opened 2026-02-06
Hadriens/fix layer availability

Farbum:
**Goal:** 
Allow training with modalities that aren't available for all windows.

**Changes:**
- Dataset skips loading non-required inputs when layers are missing
- Added skip_missing parameter to all transforms to gracefully skip missing selectors
- Added tests documenting the model batching limitation (needs separate fix, will come in a second correlated PR)
- Added tests for transforms with skip_missing and tests for Dataset to test window reading with missing non-required layers.

**Usage**
Set required: false on inputs and skip_missing: true on transforms.

### DIFF
--- a/docs/TasksAndModels.md
+++ b/docs/TasksAndModels.md
@@ -277,7 +277,7 @@ The configuration snippet below summarizes the most common options. See
 	        # Multiply ground truth values by this factor before using it for training.
 	        scale_factor: 0.1
 	        # Metric(s) to compute.
-	        # Supported: "mse", "l1", "r2".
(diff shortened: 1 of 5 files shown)


favyen2 (commented):
The main code looks good to me, thanks for adding this! For testing, is it possible to add some kind of tests for:
- Dataset itself, to initialize it and make sure that it is able to read windows that may be missing a non-required layer.
- Unit test for selector_exists.
- Unit test for one of the transforms that uses selector_exists, to make sure it works properly.

favyen2 (approved):

PR #495, opened 2026-02-07
Expected timestamps

Farbum:
**Goal**

Enable proper temporal alignment of multi-timestep data in OlmoEarth, so that missing timesteps are inserted at their correct chronological positions rather than being padded at the end.
Problem solved: When a sample has data for only some timesteps (e.g., Jan and Mar out of Jan-Apr), previously those would just be packed together at positions 0 and 1. Now they're placed at their correct temporal positions (0 and 2), with gaps (zeros + MISSING mask) at the missing positions (1 and 3).

**Changes**
1. Added expected_timestamps field to RasterImage (model_context.py)
Stores the theoretical timestamps expected for a layer, computed from window time_range and query config
2. Added compute_expected_timestamps() function (dataset.py)
Derives expected timestamps from window time_range and layer configuration
Matches the period computation logic used during actual data ingestion (aligned from end date backward)
Supports PER_PERIOD_MOSAIC mode, time_offset, and duration modifiers
3. Enhanced _prepare_modality_inputs() in OlmoEarth model (model.py)
Added _find_timestamp_position() to match actual timestamps to expected positions using midpoint matching
Added _align_tensor_to_expected_timestamps() to reorder tensor data into correct temporal positions
Missing timesteps are filled with zeros and marked with MaskValue.MISSING
Data is now chronologically ordered (oldest first)
4. Tests to ensure actual/missing timestamps are properly aligned wrt expected timestamps

### DIFF
--- a/tests/unit/train/test_dataset.py
+++ b/tests/unit/train/test_dataset.py
@@ -30,9 +30,8 @@ from rslearn.train.dataset import (
     ModelDataset,
     RetryDataset,
     SplitConfig,
-    compute_expected_timestamps,
-    read_layer_time_range,
     check_window,
+    compute_expected_timestamps,
     read_data_input,
 )
 from rslearn.train.dataset_index import INDEX_DIR_NAME
@@ -896,6 +895,8 @@ def test_compute_expected_timestamps_exact_periods_returns_timestamps() -> None:
     expected_ts = compute_expected_timestamps(window, layer_config)
     assert expected_ts is not None
     assert len(expected_ts) == 2
+
+
 class TestCheckWindow:
     """Tests for check_window and CheckWindowResult."""
 


Farbum:
> will be easier to review this after #494 is merged

It is merged now

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -95,8 +95,7 @@ def __init__(
         """
         if use_legacy_timestamps:
             warnings.warn(
-                "For new projects, don't use legacy timesteps. "
-                "Support will be removed after 2026-04-01.",
+                "For new projects, don't use legacy timesteps.",
```
favyen2: What is the reason to remove this warning? The idea would be to set default for use_legacy_timestamps to True after 2026-04-01, and then eventually remove it entirely.

rslearn/train/dataset.py
```diff
@@ -430,9 +503,27 @@ def read_data_input(
                     )
             images.append(image)
             time_ranges.append(time_range)
+
+            # Compute expected_timestamps from the first layer's config
+            # (assuming all layers in the same DataInput have same temporal config)
+            if expected_timestamps is None:
+                expected_timestamps = compute_expected_timestamps(window, layer_config)
+
+        # Sort images and time_ranges chronologically by actual timestamp
```
favyen2: I'm not sure this would work. For example the user may define two layers [sentinel2_march] and [sentinel2_february] and specify a DataInput with `layers = [sentinel2_march, sentinel2_february]` and `load_all_layers = True`. They would expect the sentinel2_march image to appear before the sentinel2_february image, but this sorting would violate the configuration (the layers would be treated more as a set than an ordered list).

What is the motivation for this sorting? I thought if you use PER_PERIOD_MOSAIC with the per_period_mosaic_reverse_time_order = False, then it should be in the correct order already.

rslearn/models/olmoearth_pretrain/model.py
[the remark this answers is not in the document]
Farbum: Done

PR #501, opened 2026-02-10
Don't apply harmonization for Sentinel-2 L2A from AWS COGS.

favyen2:
It looks like the data in s3://sentinel-cogs/ already applies the "harmonization" step to undo the changes in processing baseline >=4.00 that increases values by 1000 to highlight darker regions (harmonization undoes that to make the newly processed scenes consistent with the scenes processed before that version).


tests/online/data_sources/test_aws_sentinel2_element84.py
```diff
@@ -68,61 +70,32 @@ def test_materialize_all_bands(tmp_path: pathlib.Path, seattle2020: STGeometry)
...
-    stac_id = "S2B_56KKE_20180101_0_L2A"
-    item = harmonized_data_source.get_item_by_name(stac_id)
-    assert not item.properties["earthsearch:boa_offset_applied"]
+    To do so, we use a known scene present in both AWS and Planetary Computer with the
+    same processing baseline, and make sure that when we materialize it
+    """
+    # Item names for the known scene. The item name is unique to each data source since
+    # neither source uses the standard Sentinel-2 scene ID.
+    aws_item_name = "S2B_10TET_20260124_0_L2A"
+    pc_item_name = "S2B_MSIL2A_20260124T191559_R056_T10TET_20260124T211343"
```
APatrickJ: Based on the discussion in https://github.com/sertit/eoreader/discussions/120 it might be worth having a second test (or parameterizing this test with two sets of AWS/PC items) for pre-Jan 25, 2022

Issue #500, opened 2026-02-10
[FR] soildb 30m data source


robmarkcole:
30m is improvement over soilgrids 250m 

- https://github.com/openlandmap/soildb
- https://essd.copernicus.org/articles/18/989/2026/
- https://www.youtube.com/watch?v=443HBf1GIGI

PR #502, opened 2026-02-11
Add soildb data source

robmarkcole:
Address https://github.com/allenai/rslearn/issues/500

### DIFF
--- a/rslearn/data_sources/soildb.py
+++ b/rslearn/data_sources/soildb.py
@@ -219,7 +219,9 @@ class SoilDB(DirectMaterializeDataSource[SourceItem]):
                             for chunk in r.iter_content(chunk_size=1024 * 1024):
                                 f.write(chunk)
 
-                    tile_store.write_raster_file(item.name, band_names, UPath(local_fname))
+                    tile_store.write_raster_file(
+                        item.name, band_names, UPath(local_fname)
+                    )
 
     # --- STAC loading helpers ---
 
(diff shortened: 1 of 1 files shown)


tests/integration/data_sources/test_soildb.py
```diff
@@ -0,0 +1,184 @@
...
+        },
+    }
+    collection = {
+        "type": "Collection",
+        "stac_version": "1.0.0",
+        "id": collection_id,
+        "links": [{"rel": "item", "href": "./item.json", "type": "application/json"}],
+    }
+
+    httpserver.expect_request("/collection.json", method="GET").respond_with_json(
```
robmarkcole: Test like this?

rslearn/data_sources/soildb.py
```diff
@@ -0,0 +1,364 @@
...
+            raise ValueError(
+                f"collection {self.collection_id!r} requires asset_key to be set explicitly"
+            )
+
+        if candidates:
+            for k in candidates:
+                if k in assets:
+                    return k
+
+        if self.collection_id in SOILDB_DEFAULT_ASSET_KEY_CANDIDATES:
```
favyen2: I don't think we need these two cases (no default asset_key available for collection vs unknown collection), maybe we can just include collections for which we want to have default asset keys in SOLIDB_DEFAULT_ASSET_KEY_CANDIDATES and then have one error here that says we don't have a default asset key for that collection.

PR #509, opened 2026-02-13
Add focal loss support to SegmentationHead

Farbum:
- Add focal_loss parameter using torchvision sigmoid_focal_loss
- Add focal_loss_alpha and focal_loss_gamma parameters
- Add loss_weights tuple to scale classification and dice losses

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -35,31 +35,8 @@ jobs:
             -e CI="true" \
             rslearn-test \
             pytest -n auto tests/unit/ tests/integration/ -vv \
-            --ignore tests/unit/models/test_molmo.py \
-            --ignore tests/unit/models/olmoearth_pretrain/
+            --ignore tests/unit/models/test_molmo.py
 
       - name: Clean up
         if: always()
         run: docker image prune -f
-
-  test-olmoearth:
-    runs-on: ubuntu-latest-m
-    steps:
-      - name: Checkout repository
-        uses: actions/checkout@v4
-
-      - name: Install uv
-        uses: astral-sh/setup-uv@v4
-
-      - name: Set up Python and venv
-        run: |
-          uv python install 3.12
-          uv venv
-
-      - name: Install dependencies
-        run: |
-          uv pip install olmoearth-pretrain -e ".[extra,dev]"
-
-      - name: Run OlmoEarth Tests
-        run: |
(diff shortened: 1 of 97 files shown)


rslearn/train/tasks/segmentation.py
```diff
@@ -367,21 +381,43 @@ def forward(
...
-            per_pixel_loss = torch.nn.functional.cross_entropy(
-                logits, labels, weight=self.weights, reduction="none"
-            )
+
+            if self.focal_loss:
+                # Convert labels to one-hot for focal loss
+                num_classes = logits.shape[1]
+                labels_one_hot = (
+                    torch.nn.functional.one_hot(labels, num_classes)
+                    .permute(0, 3, 1, 2)
```
robmarkcole: Preferable to use einops

rslearn/train/tasks/segmentation.py
```diff
@@ -310,24 +312,37 @@ class SegmentationHead(Predictor):
 
     def __init__(
         self,
-        weights: list[float] | None = None,
+        class_weights: list[float] | None = None,
         dice_loss: bool = False,
+        focal_loss: bool = False,
```
favyen2: Looks like we may add more and more losses and it will get confusing. So I think we need to define CrossEntropyLoss/FocalLoss/DiceLoss and then take a `losses: list[SegmentationHeadLoss]` or `dict[str, SegmentationHeadLoss]` here (`SegmentationHeadLoss` can be subclass of `torch.nn.Module` similar to `FeatureExtractor`, `IntermediateComponent`, etc.). It can default to cross entropy loss to preserve the existing behavior.

rslearn/train/tasks/segmentation.py
```diff
@@ -310,24 +312,37 @@ class SegmentationHead(Predictor):
...
+        focal_loss_alpha: float = 0.25,
+        focal_loss_gamma: float = 2.0,
+        loss_weights: tuple[float, float] = (1.0, 1.0),
```
favyen2: Since we return the losses as dict and don't combine them, I think it is better to handle these weights in RslearnLightningModule so it can be more general-purpose. It may be possible for this to replace the loss_weights in `MultiTaskModel` (so just handling the task name prefixing in MultiTaskModel and user can configure the loss weights across tasks via RslearnLightningModule, in addition to controlling different losses like focal loss vs dice loss weight).

favyen2:

PR #511, opened 2026-02-13
Add EmbeddingCache model component and deterministic cropping transform

favyen2:


yawenzzzz (approved):
LGTM! only two minor suggestions

PR #528, opened 2026-02-24
Piperw/vis fix

piperwolters:
There was a mismatch between the window names and the visualized layer, which made debugging window creation quite confusing.


rslearn/vis/render_raster_label.py
```diff
@@ -93,4 +93,7 @@ def read_raster_layer(
         window, bounds, layer_name, group_idx, layer_config, data_input
     )
 
-    return image_tensor.numpy().astype(np.float32)
+    array = image_tensor.numpy().astype(np.float32)  # (C, T, H, W)
```
favyen2: I think it should always return CTHW now so shouldn't need if statement here.

PR #537, opened 2026-02-27
Cleanup model config parsing

favyen2:
Add a ManagedBestLastCheckpoint that should be set when using model management. It will save best.ckpt and last.ckpt to the project directory.

### DIFF
--- a/tests/unit/train/test_prediction_writer.py
+++ b/tests/unit/train/test_prediction_writer.py
@@ -3,6 +3,7 @@
 import json
 import pathlib
 from typing import Any
+from unittest.mock import Mock
 
 import numpy as np
 import numpy.typing as npt
@@ -83,8 +84,6 @@ class MockDictionaryTask(Task):
 
     def get_metrics(self) -> MetricCollection:
         """Get metrics (not used in tests)."""
-        from torchmetrics import MetricCollection
-
         return MetricCollection({})
 
 
@@ -288,8 +287,6 @@ def test_write_raster(tmp_path: pathlib.Path) -> None:
     # probabilities.
     output = torch.zeros((2, 5, 5), dtype=torch.float32)
     # Create a mock trainer to satisfy type requirements
-    from unittest.mock import Mock
-
     mock_trainer = Mock(spec=Trainer)
     mock_trainer.datamodule = Mock()
     mock_trainer.datamodule.path = UPath(ds_path)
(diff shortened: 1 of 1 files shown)


rslearn/train/callbacks/checkpointing.py
```diff
@@ -0,0 +1,111 @@
...
+            return
+
+        current_val = current.item()
+        is_better = self._best_value is None or (
+            (self.mode == "min" and current_val < self._best_value)
+            or (self.mode == "max" and current_val > self._best_value)
+        )
+        if is_better:
+            self._best_value = current_val
+            best_path = os.path.join(self.dirpath, "best.ckpt")
```
cmwilhelm: iirc the old logic would make a checkpoint with some information baked into the filename like the epoch and step numbers. This was a bit obtuse to discover, but it was nice because you could tell when the checkpoint was written and eyeball it against w&b curves.

Is it easy to recover epoch/step numbers from the embedded state dict? Or possibly we could even have a format like `f"best_epoch={epoch_num}step={step_num}.chpt"`?

rslearn/train/callbacks/checkpointing.py
```diff
@@ -0,0 +1,111 @@
...
+
+        os.makedirs(self.dirpath, exist_ok=True)
+
+        last_path = os.path.join(self.dirpath, "last.ckpt")
+        trainer.save_checkpoint(last_path)
+        logger.info("saved checkpoint to %s", last_path)
+
+        current = trainer.callback_metrics.get(self.monitor)
+        if current is None:
+            logger.warning(
```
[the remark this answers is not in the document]
favyen2: Thanks for catching that, fixed it to raise error

rslearn/train/prediction_writer.py
```diff
@@ -216,16 +217,44 @@ def __init__(
         super().__init__(write_interval="batch")
         self.output_layer = output_layer
         self.selector = selector or []
-        ds_upath = UPath(path, **path_options or {})
+
+        # Save args for use in self._initialize, which is called from setup function.
```
cmwilhelm: I'm a little unsure of what's going on with the changes in this file. Does this result in an outward behavioral change, or is this to make the existing behavior play nicely with the modifications to the managed project hooks?

cmwilhelm (approved):

PR #540, opened 2026-03-02
Fix performance issue when listing windows in groups with many windows

favyen2:
`iter_nonhidden_subdirs` is slow due to calling `is_dir` for each window. Here we move `is_dir` check into the parallelized function (`load_window`).

@robmarkcole I can't add you as reviewer but could use your feedback on this change.

### DIFF
--- a/rslearn/dataset/storage/file.py
+++ b/rslearn/dataset/storage/file.py
@@ -24,7 +24,7 @@ from .storage import WindowStorage, WindowStorageFactory
 logger = get_logger(__name__)
 
 
-def load_window(storage: "FileWindowStorage", window_dir: UPath) -> Window | None:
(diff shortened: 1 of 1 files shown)


favyen2:

rslearn/dataset/storage/file.py
```diff
@@ -34,8 +34,14 @@ def load_window(storage: "FileWindowStorage", window_dir: UPath) -> Window:
         window_dir: the path where the window is stored.
 
     Returns:
-        the window object.
+        the window object, or None if window_dir is not a directory.
     """
+    if not window_dir.is_dir():
```
favyen2: Do we actually care to do this check here? Or is it okay to fail if there are non-hidden files in the group dir?

PR #541, opened 2026-03-03
Improve robustness in match_candidate_items_to_windows when re-projection is needed

favyen2:
- Refactor safely_reproject_within_valid_area and use it when matching items to windows.
- Call shapely.make_valid in STGeometry.to_projection to handle edge cases.

### DIFF
--- a/rslearn/utils/geometry.py
+++ b/rslearn/utils/geometry.py
@@ -317,7 +317,7 @@ class STGeometry:
 
         For geometries already in WGS84, this is a no-op.
         """
-        if self.projection.crs == CRS.from_epsg(WGS84_EPSG):
+        if self.projection.crs == WGS84_PROJECTION.crs:
             if self.projection == WGS84_PROJECTION:
                 return self
             return self.to_projection(WGS84_PROJECTION)


favyen2:

rslearn/utils/geometry.py
```diff
@@ -300,17 +295,17 @@ def is_global(self) -> bool:
...
 
-        This means that it will likely have issues during re-projections and such.
+        For geometries already in WGS84, this is a no-op.
         """
-        wgs84_bounds = self.to_projection(WGS84_PROJECTION).shp.bounds
-        if wgs84_bounds[2] - wgs84_bounds[0] > MAX_GEOMETRY_DEGREES:
-            return True
-        if wgs84_bounds[3] - wgs84_bounds[1] > MAX_GEOMETRY_DEGREES:
-            return True
-        return False
```
APatrickJ: We do use this function in Studio, though I suppose we could easily write our own version that does the same thing (and ideally we want to remove the rslearn dependency from Studio entirely anyway). But for now, maybe mark this as deprecated?

rslearn/utils/geometry.py
```diff
@@ -568,56 +577,82 @@ def split_at_antimeridian(geometry: STGeometry, epsilon: float = 1e-6) -> STGeom
...
+        using shp.bounds directly would result in a big box).
+        """
+        shp = get_valid_geom_wgs84().shp
+        parts = shp.geoms if hasattr(shp, "geoms") else [shp]
+        boxes = []
+        for part in parts:
+            b = part.bounds
+            boxes.append(
+                shapely.box(
+                    b[0] - WGS84_CLIP_BUFFER_DEGREES,
```
favyen2: If we add max(width, height) here then we would better deal with large window geometries.

rslearn/utils/geometry.py
```diff
@@ -568,56 +577,82 @@ def split_at_antimeridian(geometry: STGeometry, epsilon: float = 1e-6) -> STGeom
...
+    src_geoms: Sequence[STGeometry], valid_geom: STGeometry
+) -> Sequence[STGeometry | None]:
+    """Re-project src_geoms into the projection of valid_geom.
 
-    It will first check for compatibility in WGS84, and only proceed with re-projection
-    if the geometries intersect.
+    Unlike direct to_projection(), this clips each source geometry in WGS84 to a
+    buffered area around valid_geom before reprojecting. This minimizes distortions in
+    case valid_geom is small but src_geoms may be large. It works best if src_geoms are
+    also natively in WGS84; otherwise, there could be distortion issues re-projecting
```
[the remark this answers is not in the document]
favyen2: added a notice about it in _filter_and_project_items

favyen2:

PR #546, opened 2026-03-04
Provide the Item object to the TileStore during materialization.

### LINKED ISSUE #545: Refactor materialize to avoid the need for item caches in some data sources
Some data sources (including the StacDataSource base class) use a cache to store item metadata to avoid many queries about items to the data source during materialization. These queries arise because the TileStore API (e.g. TileStore.read_raster) receives the item name but not the item object; so it needs to convert it to item object for those functions, even though the caller (like RasterMaterializer) actually does have access to the item objects.

favyen2:
This way, when using direct materialization from the data source, the data source has access to the Item object which may provide information it needs to handle TileStore functionality like get_raster_bounds and read_raster. Currently, only the name is passed, so several data sources have to lookup the Item by name; some of these implement a cache for the Item even though really we already save it in items.json and have access to it.

Other notes:
- Since the data sources now see the item during materialization, they can return RasterArrays that include the time range directly instead of needing to handle it in `materialize.py`.
- Resolves #545

### DIFF
--- a/docs/DataSources.md
+++ b/docs/DataSources.md
@@ -63,6 +63,7 @@ detailed page with configuration options and available bands. See
 | [google_earth_engine.GoogleSatelliteEmbeddings](data_sources/google_earth_engine_GoogleSatelliteEmbeddings.md) | Google Earth Engine | Google Satellite Embedding v1, requires GEE credentials |
 | [planetary_computer.PlanetaryComputer](data_sources/planetary_computer_PlanetaryComputer.md) | Microsoft Planetary Computer | Generic collection access |
 | [planetary_computer.Sentinel3SlstrLST](data_sources/planetary_computer_Sentinel3SlstrLST.md) | Microsoft Planetary Computer | Land Surface Temperature, requires ingestion |
+| [earthdaily.Biophysical](data_sources/earthdaily_Biophysical.md) | EarthDaily | LAI/FAPAR/FCOVER, requires EarthDaily credentials |
 | [xyz_tiles.XyzTiles](data_sources/xyz_tiles_XyzTiles.md) | Any XYZ tile server | Web slippy tiles |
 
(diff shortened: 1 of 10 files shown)


favyen2:
@claude can you review the PR?

favyen2:
@APatrickJ I think this is the first step that would enable updating the olmoearth_datasets data source to make the query in `materialize(self, window, item_groups, layer_name, layer_cfg)` function and it can call `RasterMaterializer().materialize(...)` with a modified item_groups that has the signed URLs cached; then when RasterMaterializer calls back to the data source `read_raster` function, `read_raster` should get those cached URLs in the Item.

rslearn/data_sources/aws_google_satellite_embedding_v1.py
```diff
@@ -353,7 +353,8 @@ def read_raster(
 
         Overrides base class to handle band selection (the base class reads all bands).
         """
-        asset_url = self.get_asset_url(item_name, "image")
+        assert isinstance(item, GoogleSatelliteEmbeddingV1Item)
+        asset_url = self.get_asset_url(item, "image")
```
APatrickJ: Usually it's best to avoid assert in prod code, could replace with:

```suggestion
        typed_item = cast(GoogleSatelliteEmbeddingV1Item, item)
        asset_url = self.get_asset_url(typed_item, "image")
```

APatrickJ (approved):
Looks good, thanks!

rslearn/data_sources/aws_google_satellite_embedding_v1.py
favyen2: That is what Claude did originally but won't that hide error when the item is not the right type? I thought cast is only for type checking and not really going to raise any error.

favyen2:

PR #550, opened 2026-03-04
Suggest to limit GPUs if there are fewer windows than GPUs.

### LINKED ISSUE #505: Running OlmoEarth Embedding Inferences with `rslearn` without Docker (Conda/Pip Environment)
## Summary

I’m trying to compute OlmoEarth embeddings via `rslearn` predict without using Docker, but the dependency stack makes this difficult to reproduce cleanly in a standard conda/pip environment. It would be helpful to have documented support for a non-Docker setup.

## What I’m Trying to Do

Run embedding inference using the example model.yaml from the [OlmoEarth Embeddings README](https://github.com/allenai/rslearn/blob/master/docs/examples/OlmoEarthEmbeddings.md). 

I would create first an rslearn environment per the [main README](https://github.com/allenai/rslearn/blob/master/README.md)
```bash
conda create -n rslearn_test python=3.12 -y
conda activate rslearn_test
git clone https://github.com/allenai/rslearn.git
cd rslearn
pip install .[extra]
```
## My Environment
- OS: I'm running this on a Linux server running Ubuntu 22.04 LTS.  
- GPU: 2× NVIDIA RTX 4090. 
- Linux Driver: 550.xx 
- CUDA 12.4
- Conda 25.7.0

## First Error and Workaround
The install fails. Key messages from the outputs:
```
Downloading bmipy-2.0.tar.gz (23 kB)
  Installing build dependencies ... done
  Getting requirements to build wheel ... error
  error: subprocess-exited-with-error
...
      AttributeError: module 'configparser' has no attribute 'SafeConfigParser'. Did you mean: 'RawConfigParser'?
      [end of output]
note: This error originates from a subprocess, and is likely not a problem with pip.
ERROR: Failed to build 'bmipy' when getting requirements to build wheel
```

I was able to circumvent this issue by using python 3.11 and running:
```
python -m pip install -U "pip<24.1" setuptools wheel
python -m pip install "cython<3" --no-cache-dir
python -m pip install --no-build-isolation "PyYAML==5.4.1"
```

This enabled me to get an rslearn environment. 

## Second Error and Workaround
Then I went to install `olmoearth_pretrain` with pip, but then got this error:
```
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
torchvision 0.24.1 requires torch==2.9.1, but you have torch 2.7.1 which is incompatible.
```
There's a fundamental incompatibility between `rslearn[extra]` dependence on torch to 2.9.1, which conflicts with `olmoearth_pretrain` (requires torch <2.8). Installing olmoearth_pretrain after rslearn[extra] downgrades torch and breaks torchvision compatibility.

To get rslearn and olmoearth_pretrain in the same environment, I needed to stop install the [extra]. The following commands finally got these two packages in one environment:
```bash
# create and activate the conda environment
conda create -n rslearn_olmo python=3.11 -y
conda activate rslearn_olmo

# Avoid the old pytorch-lightning metadata trap
pip install "pip<24.1"

# Install a consistent torch stack FIRST
pip install \
  torch==2.7.1 \
  torchvision==0.22.1 \
  torchaudio==2.7.1

# Install olmoearth
pip install olmoearth_pretrain

# Install rslearn WITHOUT forcing extras
cd ~/code/inequality-map/rslearn
pip install -e .

# Install some other needed packages skipped by not installing [extra].
conda install ipykernel # for jupyter notebook support
conda install conda-forge::matplotlib # for visualizing the GeoTIFFs
pip install wandb # for logging the training runs to Weights and Biases
pip install -U "jsonargparse[signatures]>=4.27.7"
``` 

Finally, I was able to run the following command without dependency errors:
`rslearn model predict --config model.yaml`

## Final Error in Running Model Inference:
After making it out of dependency hell, I hit an error with the model config since I only have 1 window but 2 GPUs available:
```
[/home/isaiah/miniconda3/envs/rslearnoe/lib/python3.11/site-packages/lightning/py](https://vscode-remote+ssh-002dremote-002bbplml.vscode-resource.vscode-cdn.net/home/isaiah/miniconda3/envs/rslearnoe/lib/python3.11/site-packages/lightning/py)
torch/loops/prediction_loop.py:307: Couldn't infer the batch indices fetched 
from your dataloader: `DataLoader`
...
[rank: 1] Child process with PID 2593003 terminated with code 1. Forcefully terminating all other processes to avoid zombies 🧟
```

I was able to fix this with extra runtime parameters. The final command that worked for me:
```
rslearn model predict \
    --config ./dataset/model.yaml \
    --data.init_args.path ./dataset \
    --trainer.devices 1 \
    --trainer.strategy auto
```

Finally, I got my embeddings for the demo window!
Image

## Conclusion
I think it would be great to make the OlmoEarth Embeddings tutorial runnable without Docker. Some changes are needed to make that possible.

favyen2:
Otherwise it will cause issues, when using load_all_crops=true.

Resolves #505.

### DIFF
--- a/docs/DatasetFormat.md
+++ b/docs/DatasetFormat.md
@@ -120,6 +120,9 @@ specifies which items in the data source matched with the window. Here is an exa
           }
         }
       ],
+    ],
+    "group_time_ranges": [
+      ["2020-09-01T00:00:00+00:00", "2020-09-15T00:00:00+00:00"]
     ]
   }
 ]
@@ -132,6 +135,8 @@ layer name, along with the item groups that matched. `item_groups` is a serializ
 be merged/mosaicked together to form one raster or vector file for the window. If there
 are multiple sub-lists, it typically corresponds to multi-temporal data, and each one
 will result in a different raster or vector file after the data is materialized.
+`group_time_ranges` is optional metadata parallel to `serialized_item_groups`; when
+present, it stores the exact request time range used for each item group.
 
 Materialization will use the first item group in `item_groups` to populate
 `layers/LAYER_NAME`, the second to populate `layers/LAYER_NAME.1`, and so on.
--- a/docs/dataset_config/DataSourceConfig.md
+++ b/docs/dataset_config/DataSourceConfig.md
@@ -49,7 +49,8 @@ options beyond those detailed in [LayerConfig](./LayerConfig.md):
   // It is one of "nearest", "bilinear" (default), "cubic", "cubic_spline".
   "resampling_method": "bilinear",
   // The compositing method to use, to handle raster item groups with more than one item.
-  // It can be FIRST_VALID (default), MEAN, MEDIAN, or SPATIAL_MOSAIC_TEMPORAL_STACK.
+  // It can be FIRST_VALID (default), MEAN, MEDIAN,
+  // SPATIAL_MOSAIC_TEMPORAL_STACK, TEMPORAL_MEAN, TEMPORAL_MAX, or TEMPORAL_MIN.
   "compositing_method": "FIRST_VALID"
 }
 ```
--- a/rslearn/config/dataset.py
+++ b/rslearn/config/dataset.py
@@ -588,6 +588,15 @@ class CompositingMethod(StrEnum):
     clipped to the window time range.
     """
 
+    TEMPORAL_MEAN = "TEMPORAL_MEAN"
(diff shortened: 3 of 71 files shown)


yawenzzzz (approved):
LGTM!

PR #598, opened 2026-04-08
Adds Scl compositor and separates docs

robmarkcole:
- Added a new Sentinel-2 SCL-based ranking compositor
- Added unit coverage for SCL ranking behavior
- Added config-instantiation coverage for the new compositor class path
- Docs were restructured so compositors are standalone (similar to data sources)

## Comparison with OmniCloudMask
OmniCloudMask may be struggling due to the small patch size & lack of context. Created https://github.com/DPIRD-DMA/OmniCloudMask/issues/32

sentinel2_scl_vs_omnicloudmask_broads
sentinel2_scl_vs_omnicloudmask_cairngorms
sentinel2_scl_vs_omnicloudmask_ely
sentinel2_scl_vs_omnicloudmask_kerry
sentinel2_scl_vs_omnicloudmask_lakedistrict
sentinel2_scl_vs_omnicloudmask_lincolnwolds
sentinel2_scl_vs_omnicloudmask_nymoors
sentinel2_scl_vs_omnicloudmask_southdowns
sentinel2_scl_vs_omnicloudmask_ydales


favyen2 (commented):
This looks good to me! It includes #597 so I will take another look after that PR is merged.

rslearn/data_sources/planetary_computer.py
```diff
@@ -320,9 +320,37 @@ class Sentinel2(PlanetaryComputer):
         "B11": ["B11"],
         "B12": ["B12"],
         "B8A": ["B8A"],
+        "SCL": ["SCL"],
         "visual": ["R", "G", "B"],
     }
 
+    @staticmethod
+    def _get_compositor_scoring_bands(context: DataSourceContext) -> set[str]:
```
favyen2: I prefer to leave it up to the user to enable these bands so that we don't need to check over the compositing dictionary here. It is a bit awkward since currently they would need to add it to a materialized band set (`context.layer_config` takes priority over `assets` so specifying assets explicitly to avoid needing to materialize it would not work) but I don't think this code will be easy to maintain.

favyen2 (commented):
This looks good and thanks for also adding SCL asset option for Planetary Computer! Just had one comment about the _get_compositor_scoring_bands.

rslearn/data_sources/planetary_computer.py
robmarkcole: Removed and refined the docs

favyen2 (approved):

PR #607, opened 2026-04-14
Add forkserver preload to rslearn to speed up data loader initialization

favyen2:
Previously we had this in rslearn_projects but now we don't go through rslearn_projects when running fine-tuning. I changed the preload to be fairly generic so it should only include the core requirements along with optional requirements that are needed in lightning_cli.py anyway like wandb.


piperwolters (approved):
looks good to me!

PR #609, opened 2026-04-15
MaxPoolResize

Farbum:
Adding support for a adaptive max/mean pool transform.
Useful for binary prediction tasks where the label raster is at higher resolution than the desired prediction resolution.
eg: label is at 10m resolution but we want a 40m binary prediction resolution, maxpooling over the spatial dims allow to downscale the label on the fly.

### DIFF
--- a/rslearn/train/transforms/resize.py
+++ b/rslearn/train/transforms/resize.py
@@ -92,10 +92,9 @@ class MaxPoolResize(Transform):
             skip_missing: if True, skip selectors that don't exist in the
                 input/target dicts.
         """
-        super().__init__()
+        super().__init__(skip_missing=skip_missing)
         self.target_size = target_size
         self.selectors = selectors
-        self.skip_missing = skip_missing
 
     def apply_max_pool_resize(self, image: RasterImage) -> RasterImage:
         """Apply adaptive max-pool resizing on the specified image.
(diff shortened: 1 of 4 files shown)


rslearn/train/transforms/resize.py
```diff
@@ -67,3 +68,66 @@ def forward(
         """
         self.apply_fn(self.apply_resize, input_dict, target_dict, self.selectors)
         return input_dict, target_dict
+
+
+class MaxPoolResize(Transform):
```
favyen2: I would rename this AdaptivePool or AdaptivePooling to match both the torch name and the existing GlobalPooling transform. The filename should match the class name, `adaptive_pooling.py`. Can we support both mean and max pooling here?

rslearn/train/transforms/resize.py
```diff
@@ -67,3 +68,66 @@ def forward(
...
+    """Resizes inputs to a target size using adaptive max pooling.
+
+    Unlike Resize (which uses interpolation), this uses max pooling over each
+    spatial region.  This is useful for binary label masks: if *any* pixel in a
```
favyen2: Binary label masks should probably use majority voting, not max or mean pooling. Also the docstring focuses too much on resizing, but usually adaptive pooling is not very related to that. I think it's sufficient to just say adaptive pooling here since all it does is wrap the torch function.

Farbum: > updated the docstring

rslearn/train/transforms/resize.py
Farbum: Makes sense, moved it to its own transform

favyen2 (approved):

PR #612, opened 2026-04-16
Add global raster bounds handling and tests for CHELSADaily data source

### LINKED ISSUE #610: [BUG] CHELSADaily global item bounds can collapse in local CRSs, causing materialization to fail with ValueError: No valid items found for temporal stack.
What happens:
- CHELSADaily items use a global footprint.
- In direct materialization, upstream `get_raster_bounds()` reprojects that global geometry into the target window CRS.
- For local CRSs like `EPSG:27700`, that can collapse to a degenerate bound with zero width.
- Because of that, `read_raster_window_from_tiles(...)` returns None for every CHELSA item, and `SpatialMosaicTemporalStackCompositor` raises `ValueError("No valid items found for temporal stack").`

Observed repro:
- Target projection: EPSG:27700 at 10m
- Example projected bounds from upstream `CHELSADaily.get_raster_bounds(...): (40000, -447007, 40000, 1552420)`
- Width is 0, so the item is treated as non-overlapping even though CHELSA is global.

Expected behavior:
- Global CHELSA items should remain readable for any reasonable local window CRS.
- Bounds logic should not depend on reprojecting the full global polygon into a CRS with limited validity.

Script to reproduce:
```python
"""Reproduce an rslearn CHELSA materialization failure with local-CRS windows.

This script uses upstream rslearn classes only. It demonstrates that global CHELSA
items can get degenerate projected bounds in a local CRS (for example EPSG:27700),
which then makes `read_raster_window_from_tiles(...)` return `None` and causes
`SpatialMosaicTemporalStackCompositor` to raise:

    ValueError: No valid items found for temporal stack

Usage:
    uv run python scripts/reproduce_rslearn_chelsa_global_bounds_issue.py
"""

from __future__ import annotations

from datetime import UTC, datetime

from rasterio.crs import CRS
from rasterio.enums import Resampling

from rslearn.data_sources.chelsa import CHELSADaily, CHELSADailyItem
from rslearn.dataset.compositing import SpatialMosaicTemporalStackCompositor
from rslearn.dataset.tile_utils import read_raster_window_from_tiles
from rslearn.tile_stores.tile_store import TileStoreWithLayer
from rslearn.utils.geometry import Projection, get_global_geometry


WINDOW_PROJECTION = Projection(CRS.from_epsg(27700), 10.0, -10.0)
WINDOW_BOUNDS = (62072, -28746, 62168, -28650)
REQUEST_TIME_RANGE = (
    datetime(2019, 9, 7, tzinfo=UTC),
    datetime(2019, 10, 7, tzinfo=UTC),
)


def make_source() -> CHELSADaily:
    return CHELSADaily(
        band_names=["tas", "prec"],
        extent="global",
        start_date="1979-01-01",
        end_date="2025-08-29",
        version="V.2.1",
    )


def make_item() -> CHELSADailyItem:
    return CHELSADailyItem(
        name="chelsa_daily_20190907",
        geometry=get_global_geometry(
            (
                datetime(2019, 9, 7, tzinfo=UTC),
                datetime(2019, 9, 8, tzinfo=UTC),
            )
        ),
        item_date=datetime(2019, 9, 7, tzinfo=UTC).date(),
    )


def main() -> int:
    source = make_source()
    item = make_item()
    tile_store = TileStoreWithLayer(source, "chelsa_daily")

    print("Window projection:", WINDOW_PROJECTION)
    print("Window bounds:", WINDOW_BOUNDS)
    print("Item is global:", item.geometry.is_global())

    projected_bounds = source.get_raster_bounds(
        "chelsa_daily",
        item,
        ["tas"],
        WINDOW_PROJECTION,
    )
    print("Projected CHELSA bounds in window CRS:", projected_bounds)
    print(
        "Projected width/height:",
        projected_bounds[2] - projected_bounds[0],
        projected_bounds[3] - projected_bounds[1],
    )

    raster = read_raster_window_from_tiles(
        tile_store=tile_store,
        item=item,
        bands=["tas", "prec"],
        projection=WINDOW_PROJECTION,
        bounds=WINDOW_BOUNDS,
        nodata_vals=[0, 0],
        band_dtype="float32",
    )
    print("read_raster_window_from_tiles returned None:", raster is None)

    try:
        SpatialMosaicTemporalStackCompositor().build_composite(
            group=[item],
            nodata_vals=[0, 0],
            bands=["tas", "prec"],
            bounds=WINDOW_BOUNDS,
            band_dtype="float32",
            tile_store=tile_store,
            projection=WINDOW_PROJECTION,
            resampling_method=Resampling.bilinear,
            remapper=None,
            request_time_range=REQUEST_TIME_RANGE,
        )
    except ValueError as exc:
        print("Compositor raised:", repr(exc))
        if str(exc) == "No valid items found for temporal stack":
            print()
            print("BUG REPRODUCED")
            print(
                "Expected the global CHELSA item to be readable for this window, "
                "but upstream bounds computation collapsed the global geometry to a "
                "degenerate line in the target CRS."
            )
            return 1
        raise

    print()
    print("No failure reproduced.")
    return 0


if __name__ == "__main__":
    raise SystemExit(main())
```

robmarkcole:
Close https://github.com/allenai/rslearn/issues/610

### DIFF
--- a/rslearn/data_sources/direct_materialize_data_source.py
+++ b/rslearn/data_sources/direct_materialize_data_source.py
@@ -19,7 +19,7 @@ from rslearn.dataset import Window
 from rslearn.dataset.materialize import RasterMaterializer
 from rslearn.tile_stores import TileStore, TileStoreWithLayer
 from rslearn.utils import Feature
-from rslearn.utils.geometry import PixelBounds, Projection
+from rslearn.utils.geometry import PixelBounds, Projection, get_global_raster_bounds
 from rslearn.utils.raster_array import RasterArray, RasterMetadata
 
 
@@ -144,6 +144,10 @@ class DirectMaterializeDataSource(DataSource[ItemType], TileStore, Generic[ItemT
     def get_raster_bounds(
         self, layer_name: str, item: Item, bands: list[str], projection: Projection
     ) -> PixelBounds:
+        if item.geometry.is_global():
+            # Avoid reprojecting full-world footprints into local CRSs (e.g. UTM/BNG),
+            # which can collapse to degenerate bounds.
+            return get_global_raster_bounds(projection)
         geom = item.geometry.to_projection(projection)
         return (
             int(geom.shp.bounds[0]),
--- a/rslearn/data_sources/direct_materialize_data_source.py
+++ b/rslearn/data_sources/direct_materialize_data_source.py
@@ -19,7 +19,7 @@ from rslearn.dataset import Window
 from rslearn.dataset.materialize import RasterMaterializer
 from rslearn.tile_stores import TileStore, TileStoreWithLayer
 from rslearn.utils import Feature
-from rslearn.utils.geometry import PixelBounds, Projection, get_global_raster_bounds
+from rslearn.utils.geometry import PixelBounds, Projection
 from rslearn.utils.raster_array import RasterArray, RasterMetadata
 
 
@@ -144,10 +144,6 @@ class DirectMaterializeDataSource(DataSource[ItemType], TileStore, Generic[ItemT
     def get_raster_bounds(
         self, layer_name: str, item: Item, bands: list[str], projection: Projection
     ) -> PixelBounds:
-        if item.geometry.is_global():
-            # Avoid reprojecting full-world footprints into local CRSs (e.g. UTM/BNG),
-            # which can collapse to degenerate bounds.
-            return get_global_raster_bounds(projection)
         geom = item.geometry.to_projection(projection)
         return (
             int(geom.shp.bounds[0]),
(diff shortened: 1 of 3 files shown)


rslearn/data_sources/direct_materialize_data_source.py
```diff
@@ -144,6 +144,10 @@ def get_raster_bands(self, layer_name: str, item: Item) -> list[list[str]]:
     def get_raster_bounds(
         self, layer_name: str, item: Item, bands: list[str], projection: Projection
     ) -> PixelBounds:
+        if item.geometry.is_global():
```
favyen2: I think it is better to override `get_raster_bounds` in the CHELSADaily data source. After https://github.com/allenai/rslearn/pull/541 the intention is that most of the code should work more reliably with large items as long as the item geometries are in WGS84, but it sounds like items that are actually global could still have issues.

Alternatively, you could try using safely_reproject_within_valid_area here and see if that helps.

robmarkcole: override get_raster_bounds in the CHELSADaily data source

favyen2 (approved):

PR #614, opened 2026-04-17
add processing baseline filtering to Sentinel-2 documentation

### LINKED ISSUE #613: [Enhancement] Document query examples for Planetary Computer Sentinel-2 processing baseline filtering
`rslearn.data_sources.planetary_computer.Sentinel2` supports passing a STAC query, but the docs do not currently show that this can be used to constrain Sentinel-2 processing baseline, for example:
```json
{
  "class_path": "rslearn.data_sources.planetary_computer.Sentinel2",
  "init_args": {
    "query": {
      "s2:processing_baseline": {
        "eq": "05.10"
      }
    }
  }
}
```

This would be useful to document because Planetary Computer can return scenes from multiple processing baselines within the same AOI/TOI. In a comparison study, PC returned mixed 05.09, 05.10, and 05.11 imagery for the same search window.

A short doc note and example would help users who want:
- more reproducible scene selection
- cleaner cross-provider comparisons
- a single consistent Sentinel-2 processing baseline

```json
"query": {
  "eo:cloud_cover": { "lt": 20 },
  "s2:processing_baseline": { "eq": "05.10" }
}
```
@favyen2 I wonder also if you know what was used for the pretraining of olmo, a single baseline or a mix?

robmarkcole:
close https://github.com/allenai/rslearn/issues/613

### DIFF
--- a/docs/data_sources/planetary_computer_Sentinel2.md
+++ b/docs/data_sources/planetary_computer_Sentinel2.md
@@ -13,7 +13,8 @@ Sentinel-2 satellite images from https://planetarycomputer.microsoft.com/dataset
     // baselines (recommended), see
     // https://developers.google.com/earth-engine/datasets/catalog/COPERNICUS_S2_SR_HARMONIZED
     "harmonize": false,
-    // See rslearn.data_sources.planetary_computer.PlanetaryComputer.
+    // Optional STAC query filter. Below are examples of filtering by commonly
+    // used attributes.
     "query": null,
     "sort_by": null,
     "sort_ascending": true,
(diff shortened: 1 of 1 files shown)


docs/data_sources/planetary_computer_Sentinel2.md
```diff
@@ -225,3 +225,34 @@ got included, e.g.:
   }
 }
 ```
+
+### Processing Baseline Filtering
```
favyen2: There are a lot of STAC metadata attributes that users may in theory want to filter by, and I don't think this document will be easy to read if each one has its own section. I think it would be better to have a single section that shows a single example with filters for all of the common options, e.g. processing baseline, the satellite (S2A/S2B/S2C), `eo:cloud_cover`, MGRS tile, and any others you think might be useful. Actually these could also just go in the example at the top and change the comment from "See rslearn.data_sources.planetary_computer.PlanetaryComputer" to "Optional STAC query filter. Below are examples of filtering by commonly used attributes."

robmarkcole: moved to example

favyen2 (approved):

PR #619, opened 2026-04-22
Add scoring_resolution support to OmniCloudMaskFirstValid compositor

### LINKED ISSUE #611: [ENHANCEMENT - DRAFT] OmniCloudMaskFirstValid: score once on a canonical 20 m RGB+NIR grid when using B8A
`OmniCloudMaskFirstValid` currently scores items using the projection/bounds of the current band-set materialization pass. For multi-resolution Sentinel-2 layers, that means cloud scoring is repeated at different effective resolutions.

For a common setup like:
- red_band=B04 (10 m)
- green_band=B03 (10 m)
- nir_band=B8A (20 m)

it seems more natural to score once on a canonical 20 m grid:
- downsample red/green to 20 m
- use B8A at native 20 m
- run OmniCloudMask once on (R, G, NIR)
- reuse that item ordering for all band sets in the layer

Why this seems preferable:
- B8A is already 20 m, so 20 m is the natural common resolution
- item ordering would stay consistent across 10 m / 20 m / 40 m band sets
- avoids rescoring on very small lower-resolution windows
- better matches the idea of “choose the best scene once, then materialize all associated bands in that order”

DRAFT as seeking further steering on optimum resolution/patch size in https://github.com/DPIRD-DMA/OmniCloudMask/issues/32

robmarkcole:
Close https://github.com/allenai/rslearn/issues/611

- Implements a `scoring_resolution` parameter to control when scoring is reused across band sets.
- Kept window-level caching of scoring results so an explicit scoring resolution can significantly reduce repeated OmniCloudMask inference work.
- Updated documentation to describe the simplified API and the Sentinel-2 20.0 m accuracy/speed tradeoff.
- Added and revised unit tests to validate explicit scoring-resolution behavior, default materialization-grid scoring, and related validation/error handling.


rslearn/dataset/omni_cloud_mask.py
```diff
@@ -100,6 +189,14 @@ def _score_item(
             item, scoring_bands, tile_store
         )
         if len(needed_band_sets_and_indexes) == 0:
+            if (
```
favyen2: This check seems redundant since I think we anyway require the scoring bands.

rslearn/dataset/omni_cloud_mask.py
```diff
@@ -67,10 +78,88 @@ def __init__(
...
+            ),
+        )
+
+    def _get_scoring_grid(
+        self,
+        projection: Projection,
+        bounds: PixelBounds,
+    ) -> tuple[Projection, PixelBounds]:
+        """Get the grid on which cloud ranking should be evaluated."""
+        use_canonical_b8a_grid = self.use_canonical_b8a_20m_grid
```
favyen2: Can we use the given projection and bounds and convert to 20 m/pixel instead of starting from the window projection and bounds? This would avoid the changes to `compositing.py` and `materialize.py`.

rslearn/dataset/omni_cloud_mask.py
```diff
@@ -67,10 +78,88 @@ def __init__(
...
+                tuple[str, ...],
+                Projection,
+                PixelBounds,
+                Resampling,
+                tuple[datetime, datetime] | None,
+            ],
+            tuple[str, ...],
+        ] = {}
+
+        if self.use_canonical_b8a_20m_grid is True and self.nir_band.upper() != "B8A":
```
favyen2: Is it important that the band be B8A? Or can we just call it self.use_canonical_20m_grid? Is it because for other images like Landsat it would not work well?

robmarkcole: Enquired https://github.com/DPIRD-DMA/OmniCloudMask/issues/32#issuecomment-4303078647

rslearn/dataset/omni_cloud_mask.py
robmarkcole: Simplified the API to a single scoring_resolution argument in omni_cloud_mask.py. If scoring_resolution is unset, OmniCloudMask scores on each materialization grid. If it is set, scoring runs once on a window-level grid at that resolution and the ranking is reused across band sets.

rslearn/dataset/omni_cloud_mask.py
robmarkcole: Better to allow the user to select the resolution as there is an accuracy/speed tradeoff

robmarkcole:
@favyen2 I've simplified the approach, LMK your thoughts

favyen2 (commented):
Thanks for the changes! I like the `scoring_resolution` argument and the ability to either set it to None (use band set resolution) or to a fixed value (recommended 20 m/pixel).

This looks good except I still wonder if it can be simplified to not use the window projection/bounds in the compositor and just rely on the projection and bounds passed to `build_composite`.

The passed projection and bounds should be identical to the window projection and bounds except for a resolution adjustment based on the configured zoom offset. In `materialize.py`, it is computed like this:

```python
            projection, bounds = band_cfg.get_final_projection_and_bounds(
                window.projection, window.bounds
            )
```

robmarkcole:
@favyen2 Not if we want one stable ranking reused across band sets. The projection/bounds passed to `build_composite` are band-set-specific and can be shifted by `zoom_offset` or `spatial_size` rounding, so deriving scoring from them can produce different grids, different rankings, and less cache reuse within the same window.

favyen2:
I still feel like the `prepare_for_window` API is not quite clean. But I think it could make sense to refactor the compositor in the future to process all of the band sets together. This way the compositor can make sure it processes each band set in a consistent way. It may make sense to have a convenience wrapper that iterates over the band sets so the implementing subclass can just worry about per-band-set processing, for cases where consistency across band sets is not a concern.

In that case, for now, assuming the refactor is not done here, I feel like it would be fine to either (a) ignore the rounding error since it is probably negligible and make the resolution a required argument defaulting to 20 m/pixel; or (b) remove the cache so it always recomputes it for each band set, and if scoring_resolution is None then it uses the band set resolution.

What do you think? If the API change to have compositor process all band sets together sounds reasonable to you then you could also try to implement that.

favyen2:
I think it would also be okay to simply add the window projection and bounds to the `build_composite` function. In the cache, it can include those in the cache key; this shouldn't be an issue since anyway we re-initialize the compositor for each window. This would require modifying all of the compositors. I don't like this quite as much as the other alternatives (bigger refactor of compositor, or changing OmniCloudMaskFirstValid to fit into the existing API) but it feels cleaner to me than the separate `prepare_for_window` call.

robmarkcole:
@favyen2 

- Refactored raster compositing so compositors can process all band sets for a window/group together via a new `whole-window build_composites(...)` path.
- Updated `OmniCloudMaskFirstValid` to use the new whole-window API instead of relying on `prepare_for_window(...)` state.

rslearn/dataset/materialize.py
```diff
@@ -163,7 +204,9 @@ def materialize(
             default_request_time_range = window.time_range
 
         compositor = layer_cfg.instantiate_compositor()
-
+        prepared_band_sets: list[
```
favyen2, line 167: Creating the `list[BandSetCompositeRequest]` is split between `build_composites` and here. Suggest to either move it to `build_composites` (pass `layer_cfg` and window object and let it deal with everything) or to remove `build_composites` and handle it all here.

rslearn/dataset/materialize.py
```diff
@@ -174,25 +217,38 @@ def materialize(
...
+            rasters = build_composites(
+                group=group,
+                compositor=compositor,
+                tile_store=tile_store,
+                layer_cfg=layer_cfg,
+                band_set_requests=[
+                    (band_cfg, projection, bounds, remapper)
+                    for band_cfg, projection, bounds, remapper, _ in prepared_band_sets
+                ],
+                window_projection=window.projection,
```
favyen2: Consider passing the entire window object.

rslearn/dataset/compositing.py
```diff
@@ -36,13 +37,72 @@
...
+
+
 class Compositor(ABC):
     """Abstract base for compositing methods.
 
     All built-in compositing methods and custom (jsonargparse-injectable) ones
     share this interface.
     """
 
+    def prepare_for_window(
```
favyen2: My understanding is we can remove this.

rslearn/dataset/compositing.py
```diff
@@ -36,13 +37,72 @@
...
+        request_time_range: tuple[datetime, datetime] | None = None,
+    ) -> list[RasterArray]:
+        """Build composites for multiple band sets in a window.
+
+        The default implementation preserves the existing per-band-set behavior by
+        iterating over ``requests`` and delegating to ``build_composite``. Custom
+        compositors can override this to enforce consistency across band sets or
+        to share expensive preprocessing work.
+        """
+        del window_projection, window_bounds
```
favyen2: I think this can be removed too.

tests/unit/dataset/test_omni_cloud_mask.py
```diff
@@ -154,3 +155,344 @@ def mock_predict(input_array: Any) -> np.ndarray:
...
+            )
+
+        assert scoring_reads == [
+            ("cloudy", coarse_projection, coarse_bounds),
+            ("clear", coarse_projection, coarse_bounds),
+            ("cloudy", base_projection, base_bounds),
+            ("clear", base_projection, base_bounds),
+        ]
+
+    def test_scoring_resolution_supports_non_b8a_sensors(self) -> None:
```
favyen2: I don't think this test is needed since OmniCloudMaskFirstValid no longer treats B8A in a special way.

rslearn/dataset/materialize.py
```diff
@@ -174,25 +217,38 @@ def materialize(
...
+            if len(rasters) != len(prepared_band_sets):
+                raise ValueError(
+                    "compositor.build_composites must return one raster per band set"
                 )
 
+            for (band_cfg, projection, bounds, _, raster_format), raster in zip(
+                prepared_band_sets, rasters, strict=True
+            ):
                 raster_format.encode_raster(
                     window.get_raster_dir(layer_name, band_cfg.bands, group_id),
```
favyen2, line 226: If `Compositor.build_composites` is changed to pass the window object to the compositor, consider moving the `encode_raster` call into the compositor (in which case the compositor doesn't return anything). I like the compositor returning the rasters in some ways, but the advantage of encoding within the compositor is that we can avoid having all of the rasters in memory for most compositors (all the current ones).

rslearn/dataset/materialize.py
robmarkcole: BandSetCompositeRequest creation now happens entirely inside RasterMaterializer.materialize(...)

rslearn/dataset/materialize.py
robmarkcole: Changed the whole-window compositor API to accept the full Window object instead of separate window_projection / window_bounds.

rslearn/dataset/compositing.py
robmarkcole: Removed prepare_for_window(...)

rslearn/dataset/compositing.py
robmarkcole: Removed the unused del window

tests/unit/dataset/test_omni_cloud_mask.py
robmarkcole: removed

rslearn/dataset/materialize.py
robmarkcole: Implemented an iterator approach for whole-window compositing.

- `Compositor.build_composites(...)` now yields rasters instead of returning a list in compositing.py.
- `RasterMaterializer` now consumes that iterator and calls `encode_raster(...)` as each raster is produced in materialize.py.
- `OmniCloudMaskFirstValid.build_composites(...)` now streams results instead of accumulating them first in omni_cloud_mask.py.
- Updated the Omni tests to explicitly consume the iterator in test_omni_cloud_mask.py.

Why:

- It keeps compositors focused on compositing rather than file encoding.
- It avoids holding all band-set rasters in memory at once for the common case.
- It preserves the cleaner whole-window compositor API without coupling compositors to raster formats and output paths.

rslearn/dataset/omni_cloud_mask.py
```diff
@@ -154,6 +228,88 @@ def _score_item(
...
+                    band_dtype=request.band_dtype,
+                    tile_store=tile_store,
+                    projection=request.projection,
+                    resampling_method=request.resampling_method,
+                    remapper=request.remapper,
+                    request_time_range=request_time_range,
+                )
+            )
+
     def build_composite(
```
favyen2, line 313: I think this function is not used. If it needs to be implemented due to abstract method, it could raise NotImplementedError or ValueError.

rslearn/dataset/materialize.py
favyen2: This is great!

rslearn/dataset/omni_cloud_mask.py
robmarkcole: OmniCloudMaskFirstValid.build_composite(...) now raises NotImplementedError with an explicit message.

favyen2 (approved):

Issue #622, opened 2026-04-24
[Feature Request] Add NASA HLS v2.0 (HLSS30 + HLSL30) as a DataSource


robmarkcole:
### Overview
Add support for **NASA Harmonized Landsat Sentinel-2 (HLS) v2.0** as a new raster datasource in rslearn. This covers a greater yearly range than the data on Planetary Computer (only 2020 onwards)

### Collections
- **HLSS30 v2.0** – Sentinel-2 based (2015–present)
- **HLSL30 v2.0** – Landsat 8/9 based (2013–present)

**Provider**: LP DAAC (NASA)  
**Bucket**: `lp-prod-protected` (AWS us-west-2)  
**Format**: Cloud Optimized GeoTIFFs (COGs)

### NASA Access Requirements
- Free **Earthdata Login (EDL)** account
- Long-lived **Bearer token** (`EARTHDATA_TOKEN`) generated at https://urs.earthdata.nasa.gov
- Temporary AWS S3 credentials fetched from:  
  `https://data.lpdaac.earthdatacloud.nasa.gov/s3credentials`
- Must run in **us-west-2** for optimal performance
- Granules discovered via **CMR API** (short names: `HLSS30`, `HLSL30`, version `2.0`)

### Why This Is Valuable
- Much longer archive than Planetary Computer (PC only has ~2020–present)
- Full global land coverage
- Excellent for long-term time series and change detection

### Recommended Implementation
- Support EDL token authentication + automatic temporary S3 credential refresh
- Use CMR or STAC for granule search (MGRS tile + date)
- Direct COG reading with `rasterio` + `boto3` (or `earthaccess` library)
- Support Sentinel-first + Landsat overwrite mosaicking
- Handle Fmask and quality bands
- Optional repair mode for partial scenes

### References
- Official HLS website: https://hls.gsfc.nasa.gov/
- LP DAAC S3 Access: https://data.lpdaac.earthdatacloud.nasa.gov/s3credentials
- earthaccess library: https://earthaccess.readthedocs.io/
- CMR STAC endpoint: https://cmr.earthdata.nasa.gov/stac/LPCLOUD

PR #623, opened 2026-04-24
Add NASA HLS v2.0 data sources and integration tests

robmarkcole:
Address https://github.com/allenai/rslearn/issues/622

### DIFF
--- a/rslearn/data_sources/nasa_hls.py
+++ b/rslearn/data_sources/nasa_hls.py
@@ -15,6 +15,7 @@ from urllib.parse import urlparse
 import boto3
 import rasterio
 import shapely
+from rasterio.errors import RasterioIOError
 from rasterio.session import AWSSession
 from typing_extensions import override
 from upath import UPath
@@ -34,6 +35,7 @@ from rslearn.utils.retry_session import create_retry_session
 from rslearn.utils.stac import StacItem
 
 logger = get_logger(__name__)
+_HTTP_URL_PROPERTY_PREFIX = "_http_url_"
 
 
 def _first_set_env(*names: str) -> str | None:
@@ -286,6 +288,13 @@ class _NasaHlsBase(DirectMaterializeDataSource[SourceItem], StacDataSource):
             properties_to_record=list(self.PROPERTIES_TO_RECORD),
         )
 
+    def _get_http_asset_url(self, item: SourceItem, asset_key: str) -> str | None:
+        """Return the HTTPS asset URL recorded for the given item and asset key."""
+        value = item.properties.get(f"{_HTTP_URL_PROPERTY_PREFIX}{asset_key}")
+        if isinstance(value, str):
+            return value
+        return None
+
     @contextmanager
     def _rasterio_env(self, asset_url: str) -> Iterator[None]:
         """Create a rasterio environment for the given asset URL."""
@@ -294,6 +303,29 @@ class _NasaHlsBase(DirectMaterializeDataSource[SourceItem], StacDataSource):
                 yield
             return
         yield
+
+    def _read_raster_from_local_copy(
+        self,
+        asset_url: str,
+        projection: Any,
+        bounds: Any,
+        resampling: Any,
+    ) -> tuple[Any, float | None]:
+        """Download an asset locally, then read it with rasterio."""
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            local_fname = os.path.join(tmp_dir, "asset.tif")
+            self._download_asset(asset_url, local_fname)
+            return super()._read_raster_from_url(
+                f"file://{local_fname}", projection, bounds, resampling
+            )
+
+    def _get_nodata_from_local_copy(self, asset_url: str) -> float | None:
+        """Download an asset locally to inspect its nodata metadata."""
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            local_fname = os.path.join(tmp_dir, "asset.tif")
+            self._download_asset(asset_url, local_fname)
+            with rasterio.open(local_fname) as src:
+                return src.nodata
 
     def _download_asset(self, asset_url: str, local_fname: str) -> None:
         """Download an asset to a local path for ingest."""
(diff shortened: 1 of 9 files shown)


rslearn/data_sources/nasa_hls.py
```diff
@@ -0,0 +1,819 @@
...
+    REFRESH_BUFFER = timedelta(minutes=5)
+
+    def __init__(
+        self,
+        earthdata_token: str | None,
+        s3_credentials_url: str,
+        aws_region: str,
+        timeout: timedelta,
+    ) -> None:
+        self.earthdata_token = earthdata_token or _first_set_env(
```
favyen2: I would prefer it to use one name for the environment variable instead of supporting multiple options.

rslearn/data_sources/nasa_hls.py
```diff
@@ -0,0 +1,819 @@
...
+    @override
+    def get_item_by_name(self, name: str) -> SourceItem:
+        stac_items = self.client.search(ids=[name], collections=self.collection_names)
+        if len(stac_items) == 0:
+            raise ValueError(
+                f"Item {name} not found in collections {self.collection_names}"
+            )
+        if len(stac_items) > 1:
+            raise ValueError(f"Multiple items found for ID {name}")
+        return self._stac_item_to_item(stac_items[0])
```
favyen2: Is the override mainly to allow for list of collection names? It might make sense to modify the base class to accept collection name list instead of needing to handle it here.

rslearn/data_sources/nasa_hls.py
```diff
@@ -0,0 +1,819 @@
...
+
+    @override
+    def get_items(
+        self, geometries: list[STGeometry], query_config: QueryConfig
+    ) -> list[list[MatchedItemGroup[SourceItem]]]:
+        groups = []
```
favyen2: Similar to get_item_by_name, I think it may be cleaner to handle the multiple collection names in the base class if that's the only difference.

rslearn/data_sources/nasa_hls.py
```diff
@@ -0,0 +1,819 @@
...
+        "cirrus": "B09",
+        "swir16": "B06",
+        "swir22": "B07",
+        "fmask": "Fmask",
+        "solar_azimuth": "SAA",
+        "solar_zenith": "SZA",
+        "view_azimuth": "VAA",
+        "view_zenith": "VZA",
+    }
+    BAND_ALIASES = {
```
favyen2: I think the aliases makes the documentation more complex rather than making it easier for users. Suggest to remove from all three subclasses.

rslearn/data_sources/nasa_hls.py
```diff
@@ -0,0 +1,819 @@
...
+                        UPath(local_fname),
+                        time_range=item.geometry.time_range,
+                    )
+
+
+class Hls2S30(_NasaHlsBase):
+    """NASA LP DAAC HLS v2.0 Sentinel-2 (HLSS30) data source."""
+
+    COLLECTION_NAME = "HLSS30_2.0"
+    ASSET_KEY_TO_COMMON_NAME = {
```
favyen2: The B01/B02/etc. match what we use in the Landsat and Sentinel-2 data sources, so if that is already the asset key then we can reduce a lot of the alias handling logic by just using those band names to begin with.

rslearn/utils/stac.py
```diff
@@ -149,31 +149,46 @@ def search(
...
             next_link = None
+            next_method = "GET"
             next_request_data: dict[str, Any] = {}
             for link in data.get("links", []):
                 if "rel" not in link or link["rel"] != "next":
                     continue
-                assert link["method"] == "POST"
                 next_link = link["href"]
-                next_request_data = link["body"]
+                next_method = link.get("method", "GET").upper()
```
favyen2: I think it should default to cur_method in case it is not set.

rslearn/utils/stac.py
```diff
@@ -149,31 +149,46 @@ def search(
...
         items: list[StacItem] = []
         while True:
             logger.debug(
                 "Reading STAC items from %s with request data %s", cur_url, request_data
             )
-            response = self.session.post(url=cur_url, json=request_data)
+            if cur_method == "POST":
+                response = self.session.post(url=cur_url, json=request_data)
+            elif cur_method == "GET":
+                response = self.session.get(url=cur_url)
```
favyen2, line 163: Comment here that the request_data is not passed because this only can happen if the initial response gives us a next pagination URL that is intended to use with GET method.

favyen2:

rslearn/data_sources/nasa_hls.py
robmarkcole: use only EARTHDATA_TOKEN when earthdata_token is not passed

rslearn/data_sources/nasa_hls.py
robmarkcole: Yes. That get_item_by_name override was only there to search multiple STAC collections.

Changed stac.py so StacDataSource accepts `collection_name: str | list[str]`, normalizes it to self.collection_names, and uses that for both get_item_by_name and get_items.

Updated nasa_hls.py so combined Hls2 initializes StacDataSource with its collection list and removed the duplicate get_item_by_name override. Hls2.get_items still stays custom because it maps semantic bands differently for S30 vs L30 and filters after that mapping.

rslearn/data_sources/nasa_hls.py
robmarkcole: Removed the combined Hls2.get_items override. The shared StacDataSource.get_items now handles the full multi-collection search path and has two small hooks:

- _get_sort_key for datasource-specific sort behavior
- _should_include_item for post-conversion filtering

Hls2 now only overrides those hooks. The post-conversion filter is still needed because semantic bands map to different raw asset keys for S30 vs L30, so the normal required_assets pre-filter cannot express that cleanly.

rslearn/data_sources/nasa_hls.py
robmarkcole: Removed

rslearn/data_sources/nasa_hls.py
robmarkcole: The direct NASA HLS sources now use exact HLS asset keys only.

rslearn/utils/stac.py
robmarkcole: Updated stac.py so pagination links without method inherit cur_method instead of defaulting to GET.

rslearn/utils/stac.py
robmarkcole: comment added

rslearn/data_sources/stac.py
```diff
@@ -93,13 +93,23 @@ def __init__(
         """
         self.client = StacClient(endpoint)
         self.collection_name = collection_name
```
favyen2: Remove unused `self.collection_name`

tests/unit/data_sources/test_nasa_hls.py
```diff
@@ -0,0 +1,411 @@
...
+    with pytest.raises(ValueError, match="unsupported Hls2S30 band"):
+        Hls2S30(band_names=["red"])
+
+
+def test_hls2_l30_accepts_asset_keys() -> None:
+    data_source = Hls2L30(band_names=["B01", "B04", "B05", "Fmask"])
+    assert set(data_source.asset_bands.keys()) == {"B01", "B04", "B05", "Fmask"}
+
+
+def test_hls2_l30_rejects_common_names() -> None:
```
favyen2: Since we don't handle common names in a special way, we can remove this since it is covered by `test_hls2_rejects_unknown_band`.

favyen2 (approved):

rslearn/data_sources/stac.py
robmarkcole: removed

tests/unit/data_sources/test_nasa_hls.py
robmarkcole: removed

PR #624, opened 2026-04-24
Fix error when setting deprecated nodata_vals option

favyen2:
The ValueError for self.nodata_vals + self.nodata_value being set is raised because after serialization / de-serialization, the model is validated again with both set. We have to set `self.nodata_vals` to None when we set `self.nodata_value`.


favyen2:
@Farbum this should fix errors when using the WorldCereal configuration in olmoearth_pretrain.

piperwolters (approved):

PR #626, opened 2026-04-27
[Enhancement] Add name-based alignment workflow for EarthDaily Sentinel-2 biophysical products

### LINKED ISSUE #625: [Enhancement] Add support for selecting EarthDaily biophysical products (LAI etc) by matching them to the exact Sentinel-2 scene selected for another rslearn layer.
Note: this may apply to other data sources beyond the EarthDaily biophysical product, e.g. any product (ML mask etc) derived from specific  Sentinel-2 scenes. 

EarthDaily biophysical item IDs are derived directly from Sentinel-2 scene IDs. For example:

- Sentinel-2 scene: `S2C_31TEJ_20250424_0_L2A`
- Corresponding LAI item: `S2C_31TEJ_20250424_0_L2A_LAI`

Currently, `rslearn.data_sources.earthdaily.Biophysical` searches the LAI/FAPAR/FCOVER collections independently by window geometry/time and optional static STAC `query`. That can find biophysical items in the same time window, but it does not guarantee that the selected LAI item corresponds to the same Sentinel-2 scene selected for the Sentinel-2 layer.

## Use Case

For crop-yield modeling, I want Sentinel-2 bands and EarthDaily LAI/FAPAR/FCOVER rasters to be aligned at the source-scene level. If the Sentinel-2 layer uses scene `S2C_31TEJ_20250424_0_L2A`, the LAI layer should select `S2C_31TEJ_20250424_0_L2A_LAI`, rather than any LAI product that happens to intersect the same window/date range.

This matters because LAI/FAPAR/FCOVER are generated from specific Sentinel-2 scenes, and EarthDaily processes every Sentinel-2 scene over the area without cloud filtering.

## Requested Capability

Support a config option that lets a data source match its items against items selected for another layer.

Possible API shape:

```json
"data_source": {
  "class_path": "rslearn.data_sources.earthdaily.Biophysical",
  "init_args": {
    "variable": "lai",
    "match_source_layer": "sentinel2",
    "match_source_item_template": "{source_item_name}_LAI"
  }
}
```

Alternative designs would also work, such as:

- matching by a normalized Sentinel-2 product ID stored on both items
- allowing a data source to inspect prepared item groups from another layer
- adding a generic dependent/corresponding-item data source wrapper
- adding an EarthDaily-specific option like match_sentinel2_layer

## Expected Behavior

During rslearn dataset prepare, after the Sentinel-2 layer has selected scene items for a window, the biophysical layer should select the corresponding EarthDaily item by ID or product ID.

For example:

- selected Sentinel-2 item: S2C_31TEJ_20250424_0_L2A
- selected LAI item: S2C_31TEJ_20250424_0_L2A_LAI
- selected FAPAR item: S2C_31TEJ_20250424_0_L2A_FAPAR
- selected FCOVER item: S2C_31TEJ_20250424_0_L2A_FCOVER

If a corresponding item is missing, rslearn should either fail the layer/window or skip it according to existing missing-layer behavior.

robmarkcole:
Close https://github.com/allenai/rslearn/issues/625


favyen2:

favyen2:
I think I understand the use case, but I think it is too specific while requiring too many use cases to support at this time. Instead I would suggest either:

1. Offer a data source that provides both Sentinel-2 images and the derived biophysical products together. So an item has both image assets and the derived assets.
2. Or, include a separate script in an example that takes items from one layer and uses them to somehow make items in a different layer. It could be based on changing the item name, or (might be more general) it could be looking for an item in the second layer with a very similar timestamp. Then users can run the script for different sorts of layers, and it can be documented in the example for EarthDaily Sentinel-2 + biophysical, and can also be linked from the biophysical documentation.

robmarkcole:
I agree option 2 is the better fit. 

The new workflow keeps Sentinel-2 and biophysical products as separate EarthDaily data sources. Users prepare the dataset normally, with separate layers for sentinel2, lai, fapar, and/or fcover. After rslearn dataset prepare and before materialization, they run `docs/examples/align_item_groups_by_name.py`

The script rewrites the prepared biophysical item groups so they follow the Sentinel-2 item group order. It matches by EarthDaily’s deterministic item naming pattern:

`S2C_31TEJ_20250424_0_L2A -> S2C_31TEJ_20250424_0_L2A_LAI`

By default, the script uses: `{reference_item_name}_{target_layer_upper}`
so target layers named lai, fapar, and fcover map to _LAI, _FAPAR, and _FCOVER. Then users run rslearn dataset materialize to materialize the aligned layers.

rslearn/data_sources/data_source.py
```diff
@@ -82,6 +82,30 @@ def get_items(
         """
         raise NotImplementedError
 
+    def get_items_for_windows(
```
favyen2: Revert code changes since it is handled by the script.

favyen2 (commented):
This looks good, just need to revert changes to `data_source.py`, `manage.py`, and `test_manage.py`.

rslearn/data_sources/data_source.py
robmarkcole: reverted

favyen2 (approved):

PR #629, opened 2026-04-29
[Enhancement] Add documentation and tests for ERA5 temporal stack materialization

robmarkcole:
Address https://github.com/allenai/rslearn/issues/587

- Documented the recommended ERA5 daily time-series path using SINGLE_COMPOSITE, SPATIAL_MOSAIC_TEMPORAL_STACK, and NumpyRasterFormat.
- Added guidance for temporal aggregation via TEMPORAL_MEAN, TEMPORAL_MAX, and TEMPORAL_MIN.
- Added tests covering ERA5 materialization to one CTHW NumPy raster, including clipped temporal stacks and reducer outputs.

Validation: EarthDataHub ERA5 unit tests passed locally.

### DIFF
--- a/docs/data_sources/earthdatahub_ERA5LandDailyUTCv1.md
+++ b/docs/data_sources/earthdatahub_ERA5LandDailyUTCv1.md
@@ -80,35 +80,8 @@ To materialize temporal aggregates instead, keep the same `NumpyRasterFormat` an
 
 Each reducer first builds the same clipped spatial mosaic temporal stack, then
 reduces across the T dimension to one timestep. For example, if dataset windows are
-bi-weekly, `TEMPORAL_MEAN` writes one mean aggregate per bi-weekly window:
-
-```jsonc
-{
-  "type": "raster",
-  "compositing_method": "TEMPORAL_MEAN",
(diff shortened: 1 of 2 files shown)


favyen2 (approved):
Looks good to me, I think it is useful to document alternative recommended compositing methods and such with the data source. One suggestion would be to reduce duplication by just saying to change the SPATIAL_MOSAIC_TEMPORAL_STACK to one of those other ones instead of having the whole config again, this way it is less to read for the user (especially if more sections are added in the future, it can be hard to navigate through the document if there is a lot of redundancy). Anyway you can decide whether to change that and then merge.

favyen2:
BTW @Farbum said that the numpy raster format is much faster with many bands. Do you have the same experience, where GeotiffRasterFormat should put all the timesteps into a single GeoTIFF file, but it still ends up being much slower than the numpy file due to ending up with so many dimensions on the band axis?

If the GeoTIFF works okay then it could make sense to stick with that instead of overriding the raster format.

robmarkcole:
Removed the duplicated temporal aggregate config block. 

Not done any benchmarking yet but will soon!

PR #631, opened 2026-05-01
[Bugfix] Replaced **kwargs in Sentinel2L2A.__init__ with explicit parameters

### LINKED ISSUE #630: [Bug] 'rslearn.data_sources.earthdaily.Sentinel2L2A':     Option 'cloud_cover_max' is not accepted
On configuring the arg `cloud_cover_max` the error is raised:
```json
error: Parser key "data_source":
  Problem with given class_path 'rslearn.data_sources.earthdaily.Sentinel2L2A':
    Option 'cloud_cover_max' is not accepted
```
Since `cloud_cover_max` was never part of Sentinel2L2A's explicit signature — it only exists on the parent `EarthDaily.__init__`. Since jsonargparse only introspects explicit parameters (not **kwargs), it was never a valid config option for Sentinel2L2A

robmarkcole:
Close https://github.com/allenai/rslearn/issues/630

Also updated class naming to Sentinel2C1L2A for clarity and expanded docs to cover migration from PC -> EDA

### DIFF
--- a/docs/data_sources/earthdaily_Sentinel2C1L2A.md
+++ b/docs/data_sources/earthdaily_Sentinel2C1L2A.md
@@ -1,11 +1,12 @@
-## rslearn.data_sources.earthdaily.Sentinel2
+## rslearn.data_sources.earthdaily.Sentinel2C1L2A
 
 Sentinel-2 L2A data on [EarthDaily](https://earthdaily.com/) platform using the
 Collection 1 archive (`sentinel-2-c1-l2a`).
 
-Naming note: `earthdaily.Sentinel2` means the EarthDaily Collection 1 source. For the
-older EarthDaily `sentinel-2-l2a` collection with Planetary Computer-style asset keys
-and optional DN harmonization, use `rslearn.data_sources.earthdaily.Sentinel2L2A`.
+If you are migrating from Planetary Computer or want PC-compatible asset keys and band
+names, use [`rslearn.data_sources.earthdaily.Sentinel2L2A`](earthdaily_Sentinel2L2A.md)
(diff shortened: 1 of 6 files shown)


rslearn/data_sources/earthdaily.py
```diff
@@ -109,6 +109,10 @@ def __init__(
         query: dict[str, Any] | None = None,
         sort_by: str | None = None,
         sort_ascending: bool = True,
+        cloud_cover_threshold: float | None = None,
```
favyen2: Are both `cloud_cover_threshold` and `cloud_cover_max` needed?

docs/data_sources/earthdaily_Sentinel2C1L2A.md
```diff
@@ -1,11 +1,12 @@
-## rslearn.data_sources.earthdaily.Sentinel2
+## rslearn.data_sources.earthdaily.Sentinel2C1L2A
```
favyen2, line 1: I remember something about `sentinel-2-c1-l2a` not having all the Sentinel-2 scenes originally. Have they re-processed all of the Sentinel-2 scenes with new processing baseline now?

favyen2 (approved):

docs/data_sources/earthdaily_Sentinel2C1L2A.md
robmarkcole: No but we may process it, ongoing discussion about it

rslearn/data_sources/earthdaily.py
robmarkcole: They do exactly the same thing — one is a fallback for the other. Removed `cloud_cover_threshold`

PR #632, opened 2026-05-04
Add layer-wise learning rate decay for olmoearth_pretrain model.

favyen2:
This is implemented as a new optimizer. It is OlmoEarth-specific since it needs to know the parameter names in different layers; we could make a general-purpose one in the future if we want.

FreezeUnfreeze doesn't mesh well with the new parameter groups used by LayerDecayAdamW, so this also adds SimpleFreeze that doesn't change the parameter groups, instead it just toggles requires_grad off and back on.

The suggested approach is to use LayerDecayAdamW with default parameters (but setting encoder_prefix correctly) and NOT to do any freezing. This improves performance compared to our previous recipe (i.e., freeze for 20 epochs, then unfreeze while setting backbone learning rate to 1/10 of the head learning rate).

I also simplified model.py since STModel will soon accept the fast_pass option (it just ignores it).

### DIFF
--- a/rslearn/models/olmoearth_pretrain/optimizer.py
+++ b/rslearn/models/olmoearth_pretrain/optimizer.py
@@ -51,6 +51,9 @@ class LayerDecayAdamW(OptimizerFactory):
     base_lr * layer_decay_rate ** (num_layers - i). Includes all params
     (even frozen ones) so it composes with SimpleFreeze.
 
+    num_layers should be set based on the model size (12 for tiny and base,
+    4 for nano).
+
     encoder_prefix: path from the LightningModule root to the OlmoEarth module,
         typically "model.encoder.0" when OlmoEarth is the first encoder in
         MultiTaskModel.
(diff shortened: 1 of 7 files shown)


rslearn/models/olmoearth_pretrain/optimizer.py
```diff
@@ -0,0 +1,127 @@
...
+        typically "model.encoder.0" when OlmoEarth is the first encoder in
+        MultiTaskModel.
+    """
+
+    lr: float = 0.001
+    betas: tuple[float, float] = (0.9, 0.999)
+    eps: float | None = None
+    weight_decay: float | None = None
+    layer_decay_rate: float = 0.65
+    num_layers: int = 12
```
yawenzzzz, line 67: This is just for OlmoEarth base right, have you tried LayerDecayAdamW with Tiny/Nano? also, i think it will be helpful to add an in-line comment here say that this is specifically for Base, and need to change it according to the model size. It would be even better if we can have some runtime validation that check the total layers of the model, and see if that matches with the number of layers specified here.

yawenzzzz (approved):
Looking good to me, just a comment about the num_layers

rslearn/models/olmoearth_pretrain/optimizer.py
favyen2: I started experiments with tiny/nano today morning but I didn't change the num_layers for the tiny model. I have now added validation so it will raise error if the num_layers doesn't match the layers we see based on the parameters. Also restarted the nano experiments with num_layers=4, I think tiny already has 12 layers.

PR #634, opened 2026-05-05
Add args to rslearn dataset prepare

piperwolters:
- config: specify the filename of config in dataset path rather than always assuming "config.json"
- enabled-layers: seems like specifying the layers you do want instead of the ones you don't want is as or more useful? If others agree, we could deprecate disabled-layers, or keep both


favyen2:
I think users may sometimes want to specify dataset configs that are outside the dataset directory. I suppose they could still do so by setting an absolute path here, but it would be cleaner to have it take a `config_filename: UPath | None = None`. Then I think if it is None it can default to `self.path / "config.json"` (default behavior) while if specified it would be a path relative to the user's working directory. What do you think?

piperwolters:
> I think users may sometimes want to specify dataset configs that are outside the dataset directory. I suppose they could still do so by setting an absolute path here, but it would be cleaner to have it take a `config_filename: UPath | None = None`. Then I think if it is None it can default to `self.path / "config.json"` (default behavior) while if specified it would be a path relative to the user's working directory. What do you think?

that makes sense! Just pushed an update.

rslearn/dataset/dataset.py
```diff
@@ -45,19 +45,28 @@ def __init__(
...
         disabled_layers: list[str] = [],
         dataset_config: DatasetConfig | None = None,
+        config_filepath: UPath | None = None,
+        enabled_layers: list[str] | None = None,
     ) -> None:
         """Initializes a new Dataset.
 
         Args:
             path: the root directory of the dataset
             disabled_layers: list of layers to disable
```
favyen2, line 55: I do think it makes sense to deprecate this one. Seems we generally have layers we want to enable rather than ones we want to disable.

rslearn/main.py
```diff
@@ -366,6 +366,21 @@ def add_apply_on_windows_args(parser: argparse.ArgumentParser) -> None:
...
+        type=str,
+        default=None,
+        help=(
+            "Path to dataset config JSON. Default is <root>/config.json. "
+            "Otherwise resolved like a normal path (relative to cwd unless absolute)."
+        ),
+    )
+    parser.add_argument(
+        "--enabled-layers",
+        type=parse_disabled_layers,
```
favyen2: Rename parse_disabled_layers to parse_layers or something like that.

favyen2 (approved):
looks good to me!

favyen2 (approved):

Issue #635, opened 2026-05-06
[Data source issue] CHELSA daily 30x slower than ERA5 daily


robmarkcole:
As title, using CHELSA data source is significantly slower than ERA5 daily, so much so that it is practically unusable. Creating this issue to flag this and to capture follow up investigation.

robmarkcole:
I've identified that with `ingest=true` CHELSA downloads each daily global GeoTIFF, and these are around 150MB.

robmarkcole:
resolved in pr

PR #636, opened 2026-05-06
[Enhancement] CHELSADaily to support bounding box filtering for item geometry and ingestion

robmarkcole:
Address https://github.com/allenai/rslearn/issues/635

For small AOI verified speedup using the introduced `bounds` arg:

[bounded] 0.70s, 921.0 B written
[full] 32.99s, 180.0 MiB written
bounded speedup: 46.9x faster; stored output: 204930.3x smaller

For a larger 45 window comparison, this PR speeds up chelsa by a factor of x2

## Summary

- Added `bounds` support to CHELSADaily for AOI-limited ingest from global CHELSA daily GeoTIFFs.
- Removed the `extent` arg since this is the only data available currently; CHELSA daily URLs now always use the global path.
- Updated CHELSA docs with the official dataset link, bounds guidance, and no extent config.
- Added unit coverage for bounded item geometry, outside-bounds empty groups, bounded ingest cropping, and absence of the extent arg.

### DIFF
--- a/rslearn/data_sources/chelsa.py
+++ b/rslearn/data_sources/chelsa.py
@@ -182,7 +182,9 @@ class CHELSADaily(
 
         min_lon, min_lat, max_lon, max_lat = [float(v) for v in bounds]
         if min_lon >= max_lon or min_lat >= max_lat:
-            raise ValueError("bounds must satisfy min_lon < max_lon and min_lat < max_lat")
+            raise ValueError(
+                "bounds must satisfy min_lon < max_lon and min_lat < max_lat"
+            )
         if min_lon < -180 or max_lon > 180 or min_lat < -90 or max_lat > 90:
             raise ValueError("bounds must be within WGS84 longitude/latitude limits")
         return (min_lon, min_lat, max_lon, max_lat)
--- a/rslearn/data_sources/chelsa.py
+++ b/rslearn/data_sources/chelsa.py
@@ -108,7 +108,7 @@ class CHELSADaily(
         band_names: list[str] | None = None,
         start_date: date | str = DEFAULT_START_DATE,
         end_date: date | str = DEFAULT_END_DATE,
-        bounds: list[float] | None = None,
+        bounds: tuple[float, float, float, float] | None = None,
         base_url: str = BASE_URL,
         version: str = DEFAULT_VERSION,
         timeout: timedelta = timedelta(seconds=60),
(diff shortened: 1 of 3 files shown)


rslearn/data_sources/chelsa.py
```diff
@@ -157,6 +168,24 @@ def _parse_date(value: date | str) -> date:
             return value
         return date.fromisoformat(value)
 
+    @staticmethod
```
favyen2: I don't think this validation is necessary. I would just use `self.bounds = bounds` as is. Can make it `bounds: tuple[float, float, float, float] | None = None`.

rslearn/data_sources/chelsa.py
```diff
@@ -253,9 +331,18 @@ def get_items(
...
+            matched_groups = match_candidate_items_to_window(
+                geometry, items, query_config
             )
+            if (
+                self.bounds is not None
+                and not matched_groups
+                and query_config.min_matches == 0
+            ):
+                matched_groups = [
+                    MatchedItemGroup(items=[], request_time_range=geometry.time_range)
```
favyen2: What is the reason to override the empty list from match_candidate_items_to_window with a list with empty MatchedItemGroup?

tests/unit/data_sources/test_chelsa_daily.py
```diff
@@ -41,6 +42,10 @@ def test_chelsa_daily_get_asset_url() -> None:
     )
 
 
+def test_chelsa_daily_has_no_extent_init_arg() -> None:
+    assert "extent" not in inspect.signature(CHELSADaily.__init__).parameters
```
favyen2: I don't think we need this test. It is just checking that the code looks like what it looks like and not actual functionality.

rslearn/data_sources/chelsa.py
robmarkcole: Implemented

rslearn/data_sources/chelsa.py
robmarkcole: Added a comment

tests/unit/data_sources/test_chelsa_daily.py
robmarkcole: removed

rslearn/data_sources/chelsa.py
favyen2: Is it preferred to write an empty raster for this data source when the windows are outside the configured bounds, instead of just not having an item group for it? The comment says it wants to write an empty raster but didn't seem to explain why that is desirable.

robmarkcole: Returning no item groups would leave nothing to materialize, which would make the layer appear missing/unmaterialized rather than intentionally empty. Updated the comment

favyen2: I think it makes sense for the layer to be missing if the bounds that the user has configured for the data source do not intersect the window.

robmarkcole: done

favyen2 (approved):

PR #637, opened 2026-05-06
Compute and report token count per batch

cmwilhelm:
* OlmoEarth wrapper computes this directly from TokensAndMasks, slaps
  on context dict
* lightning module copies context dict onto output metadata

### DIFF
--- a/rslearn/const.py
+++ b/rslearn/const.py
@@ -8,13 +8,10 @@ TILE_SIZE = 512
 SHAPEFILE_AUX_EXTENSIONS = [".cpg", ".dbf", ".prj", ".sbn", ".sbx", ".shx", ".txt"]
 """Extensions of potential auxiliary files to .shp file."""
 
-TOKENS_IN_BATCH_KEY = "tokens_in_batch"
-
 __all__ = (
     "WGS84_PROJECTION",
     "WGS84_EPSG",
     "WGS84_BOUNDS",
     "TILE_SIZE",
     "SHAPEFILE_AUX_EXTENSIONS",
-    "TOKENS_IN_BATCH_KEY",
 )
--- a/rslearn/const.py
+++ b/rslearn/const.py
@@ -8,10 +8,13 @@ TILE_SIZE = 512
 SHAPEFILE_AUX_EXTENSIONS = [".cpg", ".dbf", ".prj", ".sbn", ".sbx", ".shx", ".txt"]
 """Extensions of potential auxiliary files to .shp file."""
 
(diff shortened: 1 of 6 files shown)


rslearn/train/lightning_module.py
```diff
@@ -469,6 +470,11 @@ def predict_step(
             metadatas=metadatas,
         )
         model_outputs = self(context)
+
+        model_outputs.metadata[TOKENS_IN_BATCH_KEY] = context.context_dict.get(
```
favyen2: Do you think this `predict_step` can be overloaded in runner to read the key? That way the olmoearth_pretrain special case doesn't need to be handled here.

Alternatively, should we just write `model_outputs.metadata["context_dict"] = context.context_dict`?

rslearn/const.py
```diff
@@ -8,10 +8,13 @@
 SHAPEFILE_AUX_EXTENSIONS = [".cpg", ".dbf", ".prj", ".sbn", ".sbx", ".shx", ".txt"]
 """Extensions of potential auxiliary files to .shp file."""
 
+TOKENS_IN_BATCH_KEY = "tokens_in_batch"
```
favyen2: I think this fits better as a constant in `rslearn.models.olmoearth_pretrain.model`.

cmwilhelm: The reason I put it here is so I can import it in dependent projects without also transitively importing torch

rslearn/train/lightning_module.py
cmwilhelm: I understand your concern about special casing olmoearth pretrain at the level of the lightning module. I'm not sure I understand the first suggestion though.

I do like the second option though; keeps it encoder-type-specific context.

favyen2: The first one was if you are already subclassing LightningModule then you can also overload predict_step to get the context_dict and handle it.

I think the second option is fine though. Currently it seems the model components don't otherwise have a way to write to the ModelOutput.metadata.

rslearn/const.py
favyen2: You could do `rslearn/models/olmoearth_pretrain/const.py`. Although it seems like you would only get this key when running a model that depends on torch.

cmwilhelm: Yeah it turns out I didn't need this in a separate location. just shoved in the model.py alongside the OlmoEarth impl

favyen2 (approved):

rslearn/train/lightning_module.py
cmwilhelm: Took your advice to copy through the context_dict

PR #638, opened 2026-05-07
[New Data Source] Add EarthDaily Sentinel-2 EDA cloud mask data source and documentation

robmarkcole:
## Data source
Adds `rslearn.data_sources.earthdaily.Sentinel2EDACloudMask` for EarthDaily’s `sentinel-2-eda-cloud-mask` collection.

The new source exposes the first band of the `cloud-mask` asset as a single categorical cloud-mask band with values for nodata, clear, cloud, cloud shadow, and thin cloud. It supports the standard EarthDaily query/sorting/cache options, strips the second STAC band during direct reads and ingest, and includes docs plus unit coverage for config, asset inference, reading, and ingest behavior.

## Adds an EarthDaily cloud-mask clear-cover selection workflow.

The new docs example shows how to prepare both Sentinel2L2A and Sentinel2EDACloudMask layers, score prepared cloud-mask candidates by AOI clear-cover pixels, rewrite the prepared Sentinel-2 item group to the clearest related sentinel-2-l2a item, and then continue with normal rslearn ingest/materialize. A helper script is included to perform the post-prepare selection step.

### DIFF
--- a/rslearn/data_sources/earthdaily.py
+++ b/rslearn/data_sources/earthdaily.py
@@ -1218,6 +1218,156 @@ class Sentinel2L2A(EarthDaily):
                     )
 
 
+class Sentinel2EDACloudMask(EarthDaily):
+    """EarthDaily Sentinel-2 EDA cloud mask (`sentinel-2-eda-cloud-mask`) source.
+
+    The first band of the `cloud-mask` STAC asset is exposed as the rslearn band
+    `cloud-mask`: 0 nodata, 1 clear, 2 cloud, 3 cloud shadow, 4 thin cloud.
+    """
+
+    COLLECTION_NAME = "sentinel-2-eda-cloud-mask"
+
+    ASSET_BANDS = {
+        "cloud-mask": ["cloud-mask"],
+    }
+
+    def __init__(
+        self,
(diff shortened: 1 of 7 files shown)


rslearn/data_sources/earthdaily.py
```diff
@@ -1218,6 +1218,156 @@ def ingest(
...
+        projection: Projection,
+        bounds: PixelBounds,
+        resampling: Resampling = Resampling.bilinear,
+    ) -> RasterArray:
+        """Read only the first band of the cloud-mask asset."""
+        raster = super().read_raster(
+            layer_name, item, bands, projection, bounds, resampling=resampling
+        )
+        return RasterArray(
+            array=raster.array[:1, :, :, :],
```
favyen2: What are the other bands in the asset? Consider ingesting them vs stripping them here, in case they could be useful.

favyen2 (approved):
looks good to me!

rslearn/data_sources/earthdaily.py
robmarkcole: second band is a cirrus mask which is now also exposed

PR #639, opened 2026-05-07
Add window storage abstraction with per-item-group and per-layer implementations

favyen2:
The new WindowDataStorage handles the writing and reading of materialized raster and vector data on a per-window basis.

PerItemGroupStorage corresponds to the existing storage method where each item group is written as a separate file.

PerLayerStorage writes all item groups in a layer to the same file.

In theory, a future implementation could write all the layers in a window to the same file, although it would have to handle appending to the file when each successive layer is written (PerLayerStorage doesn't need to worry about this since all item groups in a layer are materialized together).

Main code changes are in `rslearn/dataset/window_data_storage/`. There are changes in `rslearn/train/dataset.py` as well. New tests in `tests/unit/dataset/window_data_storage/`. Most other changes are API updates.

Pre-existing layout (PerItemGroupStorage implementation):
```
windows/train/my_window/
├── metadata.json
├── items.json
└── layers/
    ├── sentinel2/
    │   ├── B02_B03_B04/
    │   │   └── geotiff.tif        # group 0 raster (C×1×H×W)
    ├── sentinel2.1/
    │   ├── B02_B03_B04/
    │   │   └── geotiff.tif        # group 1 raster (C×1×H×W)
    └── sentinel2.2/
        ├── B02_B03_B04/
        │   └── geotiff.tif        # group 2 raster (C×1×H×W)
```

PerLayerStorage:
```
windows/train/my_window/
├── metadata.json
├── items.json
└── layers/
    ├── sentinel2/
    │   └── B02_B03_B04/
    │       ├── geotiff.tif                # all groups packed (C×3×H×W)
    │       └── window_storage_meta.json   # {"group_timestep_counts": [1, 1, 1]}
```

This excludes completed files which would still live in the per-item-group directories. Use `rslearn.dataset.storage.sqlite.SQLiteWindowStorageFactory` to avoid it.


favyen2:

favyen2:
@claude can you review the PR?

rslearn/train/dataset.py
```diff
@@ -370,42 +385,48 @@ def read_raster_layer_for_data_input(
         # resampling. If it really is much faster to handle it via torch, then it may
         # make sense to bring back that functionality.
 
-        decode_kwargs: dict[str, Any] = {}
-        if isinstance(raster_format, NumpyRasterFormat):
```
favyen2, line 374: I removed this because there should not be special exceptions for specific raster formats here.

It seems like this was for a use case where users have a band set configured to materialize at a certain size, and they also want to read the data at that size. However, currently in the model configuration there's no way to say that the data should be read at a specific size; instead, DataInput only has the resolution factor knob. So it was pretty hacky, because it was effectively saying, even though based on the DataInput the data should be read at this resolution, if the user specified NumpyRasterFormat along with a fixed size in the band set, then we will just use that resolution. With other raster formats it would have different behavior.

I think the solution is to add spatial_size option to DataInput, but that is an issue for a future PR.

rslearn/dataset/dataset.py
```diff
@@ -102,7 +105,12 @@ def load_windows(
             names: an optional list of window names to filter loading
             kwargs: optional keyword arguments to pass to WindowStorage.get_windows.
         """
-        return self.storage.get_windows(groups=groups, names=names, **kwargs)
+        return self.storage.get_windows(
+            groups=groups,
+            names=names,
+            data_storage=self.window_data_storage,
```
favyen2: In an earlier version of this PR, we were calling `get_windows` the same as before and then iterating over the returned windows afterward to overwrite `window.data_storage` from None to `self.window_data_storage`.

I didn't like that so I changed it to the current approach. But admittedly it is still awkward to pass the WindowDataStorage into the Window(Metadata)Storage.

rslearn/dataset/window.py
```diff
@@ -267,11 +255,118 @@ def mark_layer_completed(self, layer_name: str, group_idx: int = 0) -> None:
...
+        return self.data_storage.read_vector(
+            self,
+            layer_name,
+            vector_format,
+            projection if projection is not None else self.projection,
+            bounds if bounds is not None else self.bounds,
+            group_idx=group_idx,
+        )
+
     def get_raster_dir(
```
favyen2, line 270: These functions are retained (but deprecated) so that legacy code that assumes per-item-group data storage will still work as long as PerItemGroupStorage is employed by the dataset.

rslearn/dataset/window_data_storage/per_layer.py
```diff
@@ -0,0 +1,356 @@
...
+    """Per-layer raster directory: ``layers/{layer_name}/{bandset}/``.
+
+    Note this matches the location used by ``PerItemGroupStorage`` for
+    ``group_idx=0``.
+    """
+    dirname = get_bandset_dirname(bands)
+    return window_root / "layers" / layer_name / dirname
+
+
+class _PerLayerStorageMeta:
```
favyen2, line 53: Could we avoid creating the sidecar file when all item groups are single-timestep? This way we avoid the extra file. GeotiffRasterFormat may still create an aux metadata file to store the time ranges.

favyen2: I tried it but it seems not ideal because we don't know the total number of item groups in the `read_rasters` call, so we can't validate that it indeed is likely one timestep per item group based on # item groups equaling # timesteps in the stored raster.

We could assume that no sidecar means one timestep per item group, but I don't like doing that without some validation.

There may not be a good way to expose the number of item groups to the WindowDataStorage -- it would need access to the WindowStorage for that.

rslearn/dataset/storage/storage.py
```diff
@@ -1,12 +1,15 @@
 """Abstract classes for window metadata storage."""
```
favyen2, line 1: Should we rename the metadata storage from WindowStorage -> WindowMetadataStorage?

rslearn/dataset/window_data_storage/per_layer.py
```diff
@@ -0,0 +1,356 @@
...
+        )
+        if combined.array.shape[1] != sum(meta.group_timestep_counts):
+            raise ValueError(
+                f"PerLayerStorage: combined raster T={combined.array.shape[1]} "
+                f"does not match sidecar sum {sum(meta.group_timestep_counts)}"
+            )
+        out: list[RasterArray] = []
+        for group_idx in group_idxs:
+            t_slice = meta.t_slice_for_group(group_idx)
+            sub_array = combined.array[:, t_slice, :, :]
```
APatrickJ: Minor: `read_rasters` returns numpy views into the full combined array, so the entire decoded raster stays alive until the caller drops all returned `RasterArray` references. Could add `.copy()` to the slice would let the combined array be freed earlier, at the cost of a brief ~2x peak during slicing (combined + copies coexist until `read_rasters` returns).

rslearn/dataset/storage/storage.py
APatrickJ: That seems clearer to me, but it could be done in another PR probably

rslearn/dataset/window_data_storage/per_layer.py
APatrickJ: I think it's okay being unconditional. I agree having to make the assumption based on the presence / absence of the sidecar would be awkward and seems like a smell.

rslearn/dataset/window_data_storage/per_item_group.py
```diff
@@ -0,0 +1,136 @@
...
+    folder_name = layer_name if group_idx == 0 else f"{layer_name}.{group_idx}"
+    return window_root / LAYERS_SUBDIR / folder_name
+
+
+def _per_item_group_raster_dir(
+    window_root: UPath, layer_name: str, bands: list[str], group_idx: int
+) -> UPath:
+    """Per-item-group raster directory: ``layers/{layer_name}.{group_idx}/{bandset}/``."""
+    dirname = get_bandset_dirname(bands)
+    return _per_item_group_layer_dir(window_root, layer_name, group_idx) / dirname
```
APatrickJ: Given these are used outside of this module, would it make sense to drop the leading `_`?

docs/dataset_config/WindowDataStorageConfig.md
```diff
@@ -0,0 +1,59 @@
+[Back to DatasetConfig.md](../DatasetConfig.md)
+
+## Window Data Storage
+
+The window data storage controls the on-disk layout of materialized raster and
+vector data inside each window. It is dataset-wide: all layers in a dataset
+share the same window data storage. The default stores each item group of each layer
+in a separate directory under `windows/{group_name}/{window_name}/layers/`.
+
+The default window data storage writes one directory per item group:
```
APatrickJ: Suggest putting this under its own `###` heading, maybe `PerItemGroupStorage`?

rslearn/dataset/window.py
```diff
@@ -267,11 +255,118 @@ def mark_layer_completed(self, layer_name: str, group_idx: int = 0) -> None:
         """
         self.storage.mark_layer_completed(self.group, self.name, layer_name, group_idx)
 
+    def open_layer_writer(self, layer_name: str) -> LayerWriter:
```
APatrickJ: Thought on Window surface area: this PR adds six new public members to `Window` (`data_storage`, `window_root`, `open_layer_writer`, `read_raster`, `read_rasters`, `read_vector`), all delegation wrappers over `self.data_storage`. Might be worth drawing the line here: future `WindowDataStorage` capabilities are accessed via `window.data_storage` directly rather than adding more wrappers to `Window`.

rslearn/dataset/storage/file.py
```diff
@@ -235,9 +259,8 @@ def mark_layer_completed(
     ) -> None:
         self._validate_layer_name(layer_name)
         window_path = self.get_window_root(group, name)
-        layer_dir = get_window_layer_dir(window_path, layer_name, group_idx)
-        # We assume the directory exists because the item group should be materialized
-        # before being marked completed.
+        layer_dir = _file_layer_dir(window_path, layer_name, group_idx)
+        layer_dir.mkdir(parents=True, exist_ok=True)
```
APatrickJ, line 250: With `PerLayerStorage`, this creates `layers/{layer_name}.{group_idx}/` directories that only contain the `completed` sentinel file, since the actual raster data lives at `layers/{layer_name}/{bandset}/`. Someone inspecting the window on disk would see what look like per-group data directories that contain no data.

For example, with 3 item groups:

```
layers/
├── sentinel2/
│   ├── B02_B03_B04/
│   │   ├── geotiff.tif              # packed raster (all 3 groups)
│   │   └── window_storage_meta.json
│   └── completed                    # group 0 marker
├── sentinel2.1/
│   └── completed                    # group 1 marker (no data here)
└── sentinel2.2/
    └── completed                    # group 2 marker (no data here)
```

A single completion manifest file (e.g., a JSON set of completed `(layer, group)` pairs) would avoid this.

rslearn/dataset/window_data_storage/per_layer.py
favyen2: thanks, fixed!

docs/dataset_config/WindowDataStorageConfig.md
favyen2: cleaned up the doc a bit

rslearn/dataset/window.py
favyen2: I was originally calling `window.data_storage.read_raster(window, ...)` and thought it was awkward to access it via the `window` and then also pass it. What do you think about either changing it back to that pattern, or to eliminate the `data_storage` from `Window` entirely and require the caller to get it from the `Dataset`? This also means that the window metadata storage doesn't need to know about the WindowDataStorage in `get_windows` (currently it accepts a WindowDataStorage just to attach it to each Window object).

rslearn/dataset/storage/file.py
favyen2: I think it could make sense to add a new window metadata storage or potentially consider changing FileWindowStorage in a future PR. For now I updated the doc to recommend SQLiteWindowStorage, what do you think?

rslearn/dataset/window.py
APatrickJ: Between the two options, I'd keep the delegation wrappers over `window.data_storage.read_raster(window, ...)`. The redundant `window` parameter is worse ergonomically in my opinion.

A cleaner option might be making the storage window-scoped. The dataset-level object becomes a factory with a `for_window(window) -> WindowData` method, and `Window` holds the bound instance as `self.data`. Callers write `window.data.read_raster(layer_name, bands, ...)` with no redundant parameter and no delegation wrappers on `Window`. It also means `get_windows()` doesn't need the `data_storage` parameter anymore, since `Dataset.load_windows()` can bind windows after getting them back.

Since this API hasn't shipped yet, now is the cheapest time to make that change. But it also seems entirely reasonable to merge this as-is and iterate in a follow-up.

rslearn/dataset/storage/file.py
APatrickJ: Recommending `SQLiteWindowStorage` sounds reasonable to me!

APatrickJ (approved):

rslearn/dataset/window.py
favyen2: Ah thanks that seems like a great idea! I will try it out.

favyen2: for binding in `Dataset.load_windows()`, are you thinking `Window.data` will be optional and initialized to None but then `Dataset.load_windows()` will always set it?

APatrickJ: Yes, `Window.data` would be `Optional[WindowData]` defaulting to `None`, and `Dataset.load_windows()` binds it after `get_windows()` returns.

For direct Window construction (tests, `add_windows.py`), the caller does `window.data = factory.for_window(window)`.

And then you could add a `@property` that raises a descriptive error on `None` to catch unbound windows.

favyen2:
@APatrickJ does the new implementation make sense? In https://github.com/allenai/rslearn/pull/639/commits/29f61294396b6c7d024023783def2c203730623f I did the main change and then in https://github.com/allenai/rslearn/pull/639/commits/b0990c243b0c885ad72ed33cb5d6cd7a3d42df70 I added WindowDataStorageFactory back to the constructor as an optional argument so that when callers are creating windows, they don't need to update `window._data` directly.

APatrickJ:
This looks great to me!

PR #640, opened 2026-05-08
[Enhancement] Add min_matches configuration to DataSourceConfig and update for current workflow

robmarkcole:
- Updated data source docs to describe the current `MatchedItemGroup` prepare output, including items and `request_time_range`
- Added a [Mermaid flow diagram](https://github.com/allenai/rslearn/blob/refine-data-overview-docs/docs/dataset_config/DataSourceConfig.md) showing Prepare, Ingest, and Materialize, with materialization depending on both `items.json` metadata and tile store assets.
- Documented `min_matches` in the dataset config examples and prepare-stage option list.
- Clarified DatasetFormat.md wording around `serialized_item_groups` and parallel `group_time_ranges`

### DIFF
--- a/docs/dataset_config/DataSourceConfig.md
+++ b/docs/dataset_config/DataSourceConfig.md
@@ -23,6 +23,9 @@ options beyond those detailed in [LayerConfig](./LayerConfig.md):
       "space_mode": "MOSAIC",
       // The max matches defaults to 1.
       "max_matches": 1,
+      // The min matches defaults to 0. If fewer item groups are found, the window is
+      // rejected for this layer and no item groups are returned.
+      "min_matches": 0,
       // For MOSAIC, the number of overlapping items wanted within each item group covering
       // the window (default 1). Set higher for compositing.
       "mosaic_compositing_overlaps": 1,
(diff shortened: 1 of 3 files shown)


docs/dataset_config/DataSourceConfig.md
```diff
@@ -124,11 +166,16 @@ Suppose the window time range is [2024-01-01, 2024-02-01].
...
-### Space Mode and Max Matches
+### Space Mode, Max Matches, and Min Matches
 
 The `space_mode` defines the matching strategy. It interacts with `max_matches`, which
 specifies the maximum number of item groups to produce.
 
+`min_matches` specifies the minimum number of item groups required. If matching produces
+fewer than `min_matches` groups, the window is rejected for that layer and prepare stores
+no item groups for it. The default is 0, so windows with no matching items are not
+rejected unless a data source implements stricter custom behavior.
```
favyen2: I don't think data source usually would override the behavior specified by the QueryConfig. So I think it could read "The default is 0, so windows with no matching items are not rejected unless the user configures a stricter behavior."

favyen2 (approved):

docs/dataset_config/DataSourceConfig.md
robmarkcole: changed!

PR #642, opened 2026-05-08
[Bugfix] Implement axis-aligned geometry handling for STAC intersects searches

### LINKED ISSUE #641: [Bug] StacDataSource.get_items() sends non-axis-aligned polygon to STAC intersects, causing HTTP 500 from NASA CMR
When a dataset window is in a non-WGS84 projection (e.g. EPSG:27700 British National Grid), StacDataSource.get_items() reprojects the window polygon to WGS84 and passes it verbatim as the intersects geometry in the STAC search request. Due to meridian convergence in the source projection, the reprojected polygon is a slightly rotated quadrilateral rather than an axis-aligned rectangle. NASA CMR's STAC endpoint (https://cmr.earthdata.nasa.gov/stac/LPCLOUD/search) returns HTTP 500 for any non-axis-aligned polygon in intersects, even when the polygon is geometrically valid. This affects all NASA HLS data sources (Hls2S30, Hls2L30, Hls2) and any other StacDataSource subclass queried against CMR when windows are not in WGS84.

To reproduce:

- Create a window in EPSG:27700 (or any non-WGS84 projected CRS).
- Configure a layer backed by nasa_hls.Hls2S30 (or Hls2L30 / Hls2).
- Run rslearn dataset prepare.

The prepare phase enters a retry loop, logging:
```
loglevel=WARNING logger=rslearn.dataset.manage retry() L83
Retrying after catching error in retry loop: 500 Server Error: Internal Server Error
for url: https://cmr.earthdata.nasa.gov/stac/LPCLOUD/search
```
The reprojected polygon sent to CMR looks like:
```json
{
  "type": "Polygon",
  "coordinates": [[
    [-0.5526254425357197, 52.70582728785213],
    [-0.5526849071324283, 52.70402993899932],
    [-0.5603781508704868, 52.70412359827854],
    [-0.5603190021503118, 52.70592095318912],
    [-0.5526254425357197, 52.70582728785213]
  ]]
}
```
Direct verification that the polygon type matters:
```
# Non-axis-aligned polygon → 500
curl -s -o /dev/null -w "%{http_code}" -X POST \
  https://cmr.earthdata.nasa.gov/stac/LPCLOUD/search \
  -H "Content-Type: application/json" \
  -d '{"collections":["HLSS30_2.0"],"intersects":{"type":"Polygon","coordinates":[[[-0.5526,52.7058],[-0.5527,52.7040],[-0.5604,52.7041],[-0.5603,52.7059],[-0.5526,52.7058]]]},"datetime":"2025-01-01T00:00:00Z/2025-03-31T23:59:59Z","limit":10}'
# → 500

# Axis-aligned bbox polygon → 200
curl -s -o /dev/null -w "%{http_code}" -X POST \
  https://cmr.earthdata.nasa.gov/stac/LPCLOUD/search \
  -H "Content-Type: application/json" \
  -d '{"collections":["HLSS30_2.0"],"intersects":{"type":"Polygon","coordinates":[[[-0.5604,52.7040],[-0.5526,52.7040],[-0.5526,52.7059],[-0.5604,52.7059],[-0.5604,52.7040]]]},"datetime":"2025-01-01T00:00:00Z/2025-03-31T23:59:59Z","limit":10}'
# → 200
```

Expected behavior: prepare completes without error. CMR returns matching HLS items for the window's spatial/temporal extent.

### Proposed fix

In StacDataSource.get_items(), replace wgs84_geometry.shp with wgs84_geometry.shp.envelope when building the intersects argument:

```python
# Before
stac_items = self.client.search(
    collections=self.collection_names,
    intersects=json.loads(shapely.to_geojson(wgs84_geometry.shp)),
    ...
)

# After
stac_items = self.client.search(
    collections=self.collection_names,
    intersects=json.loads(shapely.to_geojson(wgs84_geometry.shp.envelope)),
    ...
)
```

.envelope returns the minimum axis-aligned bounding box of the polygon, which is always a valid rectangle in WGS84 and accepted by CMR. No spatial accuracy is lost: match_candidate_items_to_window() performs fine-grained intersection filtering on the returned items downstream.

### Environment

- rslearn version: 0.1.3
- Affected data sources: rslearn.data_sources.nasa_hls.Hls2S30, Hls2L30, Hls2
- STAC endpoint: https://cmr.earthdata.nasa.gov/stac/LPCLOUD

robmarkcole:
Close https://github.com/allenai/rslearn/issues/641

### DIFF
--- a/rslearn/data_sources/stac.py
+++ b/rslearn/data_sources/stac.py
@@ -17,7 +17,9 @@ from rslearn.utils.stac import StacClient, StacItem
 logger = get_logger(__name__)
 
 
-def _get_search_intersects_shape(shp: shapely.Geometry) -> shapely.Geometry:
+def _get_axis_aligned_search_intersects_shape(
+    shp: shapely.Geometry,
+) -> shapely.Geometry:
     """Return an axis-aligned geometry suitable for STAC intersects searches."""
     if isinstance(shp, shapely.MultiPolygon):
         envelopes = [part.envelope for part in shp.geoms if not part.is_empty]
@@ -168,6 +170,10 @@ class StacDataSource(ItemLookupDataSource[SourceItem]):
         assert self.sort_by is not None
         return stac_item.properties[self.sort_by]
 
+    def _get_search_intersects_shape(self, shp: shapely.Geometry) -> shapely.Geometry:
+        """Return the geometry to use for STAC intersects searches."""
+        return shp
+
     def _should_include_item(self, item: SourceItem) -> bool:
         """Return whether the converted item should be included in search results."""
         return True
@@ -218,7 +224,9 @@ class StacDataSource(ItemLookupDataSource[SourceItem]):
             stac_items = self.client.search(
                 collections=self.collection_names,
                 intersects=json.loads(
-                    shapely.to_geojson(_get_search_intersects_shape(wgs84_geometry.shp))
+                    shapely.to_geojson(
+                        self._get_search_intersects_shape(wgs84_geometry.shp)
+                    )
                 ),
                 date_time=search_time_range,
                 query=self.query,
@@ -271,3 +279,11 @@ class StacDataSource(ItemLookupDataSource[SourceItem]):
     def deserialize_item(self, serialized_item: dict) -> SourceItem:
         """Deserializes an item from JSON-decoded data."""
         return SourceItem.deserialize(serialized_item)
+
+
+class AxisAlignedStacDataSource(StacDataSource):
+    """STAC data source variant for APIs requiring axis-aligned search geometry."""
+
+    def _get_search_intersects_shape(self, shp: shapely.Geometry) -> shapely.Geometry:
+        """Return an axis-aligned geometry to use for STAC intersects searches."""
+        return _get_axis_aligned_search_intersects_shape(shp)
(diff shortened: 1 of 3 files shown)


rslearn/data_sources/stac.py
```diff
@@ -206,7 +217,9 @@ def get_items(
             search_time_range = self._get_search_time_range(wgs84_geometry)
             stac_items = self.client.search(
                 collections=self.collection_names,
-                intersects=json.loads(shapely.to_geojson(wgs84_geometry.shp)),
+                intersects=json.loads(
+                    shapely.to_geojson(_get_search_intersects_shape(wgs84_geometry.shp))
```
favyen2: Would it make sense to have a subclass of this for use by data sources that involve STAC APIs that don't support non-axis-aligned geometries? Because I think if there is a large rotation, it is desirable to use the actual geometry (which should be relatively accurate for small windows; for super large windows there could be distortion that is not captured by straight edges between vertices, but most uses involve small windows), so that the identified items actually intersect the window. I guess using a bbox is okay since it will be caught later, but I don't think the limitation specific to some STAC APIs should influence the general-purpose class.

robmarkcole: Added AxisAlignedStacDataSource for APIs that require axis-aligned intersects polygons, and moved NASA HLS onto that subclass

favyen2 (approved):

PR #647, opened 2026-05-14
Remove incorrect GlobalPool/in_channels additions; document PoolingDecoder for patch regression

robmarkcole:
- Remove `GlobalPool` component (added in error; only useful when C is already the desired output size, which it never is from a raw backbone embedding)
- Remove `in_channels` projection from `RegressionHead` (projection belongs in the decoder, not the head)
- Add a usage note to `PoolingDecoder` documenting `out_channels=1` as the correct pattern

**Recommended config for OlmoEarth → regression:**
```yaml
encoder:
  - class_path: rslearn.models.olmoearth_pretrain.model.OlmoEarth
    init_args:
      model_id: OLMOEARTH_V1_NANO  # 128-channel output
decoder:
  - class_path: rslearn.models.pooling_decoder.PoolingDecoder
    init_args:
      in_channels: 128
      out_channels: 1              # pools spatially and projects Bx128 → Bx1
  - class_path: rslearn.train.tasks.regression.RegressionHead
```


rslearn/train/tasks/regression.py
```diff
@@ -265,11 +266,16 @@ def __init__(
                 requires targets to be between 0-1.
             huber_delta: delta parameter for Huber loss (only used when
                 loss_mode="huber").
+            in_channels: if set, adds a linear projection from in_channels to 1 before
```
favyen2: I think there could be a Linear component similar to Conv for this. I don't quite think the head should have the feature.

favyen2: Actually I think this use case might already be covered by PoolingDecoder. I think GlobalPool is only intended for situations where you want to keep all the embedding dimensions. So it might be more about improving documentation than adding feature to the code.

robmarkcole: Removed in_channels and updated the description in this PR

rslearn/models/global_pool.py
```diff
@@ -48,8 +53,8 @@ def forward(
             context: the model context.
 
         Returns:
-            If keep_spatial_dims=False (default): FeatureVector (BxC) suitable for
```
favyen2, line 51: I think the point is it is suitable as long as C is already your desired output channels.

So for classification head, you would need to first apply Conv or similar to get it to have channel dimension be the number of classes. Then GlobalPool would average the logits across all spatial patches.

For regression head, you would do the same but output one from the conv and then GlobalPool would average the predicted value.

I don't think GlobalPool from 768-dim embedding to ClassificationHead would work.

favyen2:
I couldn't find any uses of GlobalPool, I'm not sure what the original use case was.

robmarkcole:
@favyen2 GlobalPool looks like tech-debt then so removed in this PR

favyen2 (approved):
makes sense to me, thanks for the cleanup!

PR #650, opened 2026-05-18
Add FindStadiums and GetImageTimeSeries examples

favyen2:
- [FindStadiums](https://github.com/allenai/rslearn/blob/a22b34fa4a8592e2bb5f547c24850dd44e9de574/docs/examples/FindStadiums.md): shows how to fine-tune models when you only have point labels. Also shows how to compare OlmoEarth against other models like Clay and Prithvi.
- [GetImageTimeSeries](https://github.com/allenai/rslearn/blob/a22b34fa4a8592e2bb5f547c24850dd44e9de574/docs/examples/GetImageTimeSeries.md): shows how to use rslearn to get Sentinel-2 image series of some target locations.
- Fix bug in prithvi where positional embedding was being created on CPU for multi-temporal input


favyen2:
Depends on #633

docs/examples/FindStadiums.md
```diff
@@ -0,0 +1,605 @@
...
+    }
+  }
+}
+```
+
+The script below implements the data strategy discussed above. We skip Washington since
+we will test the model on a portion of Seattle, WA later. The windows all share the same
+90-day time range, from 1 June 2025 to 30 August 2025.
+
+```python
```
gabrieltseng, line 146: I think this script is the crux of the tutorial (how to turn a list of labels into a whole bunch of windows). 

I wonder if it deserves its own markdown file explaining all of the concepts, or at least a bunch more comments explaining the different objects being loaded from rslearn.

gabrieltseng: (I don't think this should block merging this in, but I would be keen to see this. cc @ahoweai2 )

gabrieltseng (approved):

docs/examples/FindStadiums.md
favyen2: I added more comments, especially for the rslearn objects like Dataset, Window, Projection, and STGeometry. I think some of the other tutorials that include some code like ProgrammaticWindows could also benefit from more comments.

PR #657, opened 2026-05-28
Data loading speed up

favyen2:
This resolved one issue and one bottleneck when using PerLayerStorage + NumpyRasterFormat.

- NumpyRasterFormat didn't previously support reading different bounds than the bounds of the stored image. It now supports this. The requested projection must still match the projection of the stored image (i.e., re-projection is not supported). SingleImageRasterFormat is updated as well to use the same helper function.
- The data loading speed is much faster with PerLayerStorage + NumpyRasterFormat, and `BandSetConfig.instantiate_raster_format` becomes the bottleneck. We resolve it by caching the instantiated RasterFormat.

Here are benchmark numbers:

```
config:
- A100 VM on GCP (12 cores)
- 32x32 inputs, batch_size=8, 12 timesteps, Sentinel-2 only (12 bands)
- These numbers are with OlmoEarth-v1-Base
- Showing the time for the read_raster_layer_groups_for_data_input function, also showing batches/sec for some configurations

one data loader worker, per item group storage, with RasterFormat caching

npy balanced: 0.0083s
npy ssd: 0.0088s
lzw balanced: 0.0315s
lzw ssd: 0.0315s
zstd balanced: 0.0302s
zstd ssd: 0.0296s
uncompressed balanced: 0.0297s
uncompressed ssd: 0.0286s

one data loader worker, per layer storage, with RasterFormat caching

npy balanced: 0.0018s
npy ssd: 0.0018s
lzw balanced: 0.0120s
lzw ssd: 0.0121s
zstd balanced: 0.0101s
zstd ssd: 0.0103s
uncompressed balanced: 0.0095s
uncompressed ssd: 0.0099s

one data loader worker, per layer storage, no RasterFormat caching

npy: 0.0037s, 7.10 batch/sec
lzw: 0.0147s, 5.64 batch/sec

8 data loader workers, per item group storage, with RasterFormat caching

npy: 0.0078s, 7.13 batch/sec
lzw:  0.0305s, 7.09 batch/sec
uncompressed: 0.0296s, 7.02 batch/sec

8 data loader workers, per layer storage, with RasterFormat caching

npy balanced: 0.0019s, 7.13 batch/sec
npy ssd: 0.0019s, 7.09 batch/sec
lzw balanced: 0.0116s, 7.09 batch/sec
lzw ssd: 0.0120s, 7.09 batch/sec
uncompressed balanced: 0.0096s, 7.06 batch/sec
uncompressed ssd: 0.0100s, 7.03 batch/sec
```

### DIFF
--- a/rslearn/utils/array.py
+++ b/rslearn/utils/array.py
@@ -77,8 +77,8 @@ def copy_spatial_array(
         dst_offset: the (col, row) position of the top-left pixel of dst in the coordinate
             system.
     """
-    if len(src.shape) < 2:
-        raise ValueError(f"src must be at least 2-D, got shape {src.shape}")
+    if len(src.shape) < 2 or len(dst.shape) < 2:
+        raise ValueError(f"src and dst must be at least 2-D, got shape {src.shape}")
 
     src_height, src_width = src.shape[-2:]
     dst_height, dst_width = dst.shape[-2:]
--- a/rslearn/utils/array.py
+++ b/rslearn/utils/array.py
@@ -66,14 +66,20 @@ def copy_spatial_array(
     array will be copied, and other parts of the destination array will not be
     overwritten.
 
+    Supports arrays with any number of dimensions >= 2. The last two dimensions are
+    treated as (H, W); any leading dimensions (e.g. C, or C and T) are preserved.
+
     Args:
-        src: the source array (HW or CHW).
-        dst: the destination array (HW or CHW).
+        src: the source array (``...HW``).
+        dst: the destination array (``...HW``).
         src_offset: the (col, row) position of the top-left pixel of src in the coordinate
             system.
         dst_offset: the (col, row) position of the top-left pixel of dst in the coordinate
             system.
     """
+    if len(src.shape) < 2:
+        raise ValueError(f"src must be at least 2-D, got shape {src.shape}")
+
     src_height, src_width = src.shape[-2:]
     dst_height, dst_width = dst.shape[-2:]
     # The top-left position within src that intersects with dst.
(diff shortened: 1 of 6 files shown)


rslearn/utils/array.py
```diff
@@ -87,23 +93,12 @@ def copy_spatial_array(
...
-            dst_row_offset : dst_row_offset + row_overlap,
-            dst_col_offset : dst_col_offset + col_overlap,
-        ] = src[
-            :,
-            src_row_offset : src_row_offset + row_overlap,
-            src_col_offset : src_col_offset + col_overlap,
-        ]
-    else:
-        raise ValueError(f"Unsupported src shape: {src.shape}")
+    dst[
```
cmwilhelm, line 100: Was this solving a specific issue or just generalizing the operation to any dims > 2?

rslearn/utils/raster_format.py
```diff
@@ -766,16 +819,6 @@ def decode_raster(
         with (path / METADATA_FNAME).open() as f:
             image_metadata = SingleImageRasterMetadata.model_validate_json(f.read())
 
-        # If the projection key is set, verify that it matches the requested projection
-        # since SingleImageRasterFormat currently does not support re-projecting.
```
cmwilhelm, line 770: Why are we removing this validation? Is it just not supportable?

cmwilhelm: Oh wait, you just moved it out nm.

tests/unit/utils/test_raster_format.py
```diff
@@ -385,15 +385,62 @@ def test_files_created(self, tmp_path: pathlib.Path) -> None:
         assert (path / "data.npy").exists()
         assert (path / "metadata.json").exists()
 
-    def test_bounds_mismatch_raises(self, tmp_path: pathlib.Path) -> None:
-        """Decoding with different bounds should raise ValueError."""
+    def test_bounds_subset(self, tmp_path: pathlib.Path) -> None:
```
cmwilhelm: I might appreciate a tiny "integration" test in this file that writes both a Geotiff and Numpy raster format to disk, and performs equivalent decode operations (with the custom bounds) to ensure the data that's decoded is identical.

cmwilhelm (approved):
I am not equipped to review all the numpy operations but I did have one small suggestion about an additional test to include.

tests/unit/utils/test_raster_format.py
favyen2: I consolidated the tests that write data and read it back into TestRasterFormatRoundTrip. Is that what you were thinking? I think Geotiff/Numpy had some tests before but they were writing constant value across the raster, the new consolidated tests use np.arange. And now they also apply to all the raster formats instead of just having specialized tests for those two.

As part of it I also updated ImageTileRasterFormat to have correct nodata handling.

rslearn/utils/array.py
favyen2: Just generalizing to > 2 dims so we can apply it to the CTHW arrays that RasterFormat operates on.

favyen2:

tests/unit/utils/test_raster_format.py
cmwilhelm: Oh yeah the roundtripping strategy over all the variants is better than what I suggested and captures what I was looking for. Nice, thanks.

PR #658, opened 2026-06-01
Add GLO30 (read from AWS S3 bucket)

yawenzzzz:
Adding GLO30 dataset, besides elevation we also compute aspect and slope


docs/data_sources/aws_glo30_CopernicusGLO30.md
```diff
@@ -0,0 +1,93 @@
...
+  "ingest": true
+}
+```
+
+### Available Bands
+
+The data source should be configured with a single band set containing one or more of:
+
+- `elevation` — raw DEM value in meters
+- `slope` — terrain slope in degrees [0, 90)
```
favyen2, line 37: We materialize raw bands for most data sources and compute derived bands on the fly, could we do the same here? For example slope/aspect could be computed with a transform during model training/inference based on the elevation values.

rslearn/data_sources/aws_glo30.py
```diff
@@ -0,0 +1,410 @@
...
+                "expected mosaic with max_matches=1 for the query configuration"
+            )
+        if query_config.min_matches != 0:
+            raise ValueError(
+                "min_matches is not supported for CopernicusGLO30; set min_matches=0"
+            )
+
+        groups = []
+        for geometry in geometries:
+            wgs84_geometry = geometry.to_wgs84()
```
favyen2, line 282: worth double checking whether it makes sense to call the antimeridian splitting code here

rslearn/data_sources/aws_glo30.py
```diff
@@ -0,0 +1,410 @@
...
+            logger.debug(f"Got {len(items)} GLO-30 items for geometry")
+            groups.append([MatchedItemGroup(items, geometry.time_range)])
+
+        return groups
+
+    def deserialize_item(self, serialized_item: dict) -> Item:
+        """Deserializes an item from JSON-decoded data."""
+        return Item.deserialize(serialized_item)
+
+    def ingest(
```
favyen2, line 310: Since they are COGs can we also support direct materialization in this data source? Or do they need to be ingested and converted first?

favyen2:

rslearn/data_sources/aws_glo30.py
favyen2: oh I see, `to_wgs84` already handles this!

docs/data_sources/aws_glo30_CopernicusGLO30.md
yawenzzzz: Sounds good, moved slope/aspect computation to Transform.

rslearn/data_sources/aws_glo30.py
yawenzzzz: Yeah, now only for direct materialization. The previous ingestion is mainly for computing aspect/slope, now they moved to Transform.

rslearn/train/transforms/terrain.py
```diff
@@ -0,0 +1,167 @@
...
+                aspect = torch.where(
+                    slope < FLAT_SLOPE_EPSILON,
+                    torch.full_like(aspect, self.flat_aspect_value),
+                    aspect,
+                )
+                band_arrays[ASPECT_BAND] = aspect
+
+        # Central differences do not read the center pixel, so a pixel with unknown
+        # elevation would otherwise get a slope derived purely from its neighbors.
+        nan_mask = torch.isnan(dem)
```
favyen2, line 120: Document in class docstring that NaN is assumed to be nodata value.

favyen2 (approved):

rslearn/train/transforms/terrain.py
yawenzzzz: Fixed.

PR #662, opened 2026-06-04
Add MapAllCropsDataset for memory-efficient all-crops prediction -- Sentinel1 Window OOM Fix

rpw1134:
Add a map-style all-crops dataset that reads each crop's bounds from disk on demand, as the map-style counterpart to IterableAllCropsDataset. It enumerates every crop (exposing __len__/__getitem__) and, unlike InMemoryAllCropsDataset, does not cache windows, so peak memory is one crop regardless of window size.

Select it via the use_map_all_crops_dataset flag on RslearnDataModule, akin to use_in_memory_dataset.

### DIFF
--- a/rslearn/train/all_crops_dataset.py
+++ b/rslearn/train/all_crops_dataset.py
@@ -1,6 +1,7 @@
 """Wrapper around ModelDataset to load all crops in a window."""
 
 import itertools
+import random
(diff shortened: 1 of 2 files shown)


favyen2 (approved):
looks good to me, thanks for moving this here

PR #665, opened 2026-06-08
Add torchmetrics[detection] and jsonargparse[signatures] as core requirements

favyen2:
They were previously only brought in via some extra dependencies so now even rslearn[extra] is insufficient. jsonargparse[signatures] is needed for the LightningCLI while torchmetrics[detection] is needed for DetectionTask.


pyproject.toml
```diff
@@ -13,7 +13,7 @@ dependencies = [
     "fiona>=1.10",
     "flask>=3.0.0",
     "fsspec>=2025.10.0", # this is used both directly and indirectly (via universal_pathlib) in our code
-    "jsonargparse>=4.35.0",
+    "jsonargparse[signatures]>=4.35.0",
```
robmarkcole, line 16: Note https://github.com/torchgeo/torchgeo/blob/887c48fca940a6e0050da3af33f6aa836b11c4c0/pyproject.toml#L45 and https://github.com/Lightning-AI/pytorch-lightning/pull/21747

joshhvulcan (approved):

rpw1134 (approved):

pyproject.toml
favyen2: Are you suggesting to add `jsonargparse[jsonnet]` and not just `jsonargparse[signatures]`? I checked a bit about it and I think since we don't use jsonnet we may not need this but if you run into issue where it says it is needed even when not using `.jsonnet` config files then open an issue / PR.

PR #666, opened 2026-06-09
Add interactive osm map to the vis server with points/polygons overlaid

piperwolters:
Add interactive map with the ROI highlighted (just a bbox for points and polygons) to the vis server for each window visualized.

Screenshot 2026-06-09 at 3 03 52 PM
Screenshot 2026-06-09 at 3 04 15 PM


rslearn/data_sources/openstreetmap.py
```diff
@@ -32,20 +33,50 @@ class FeatureType(Enum):
...
+    return [_parse_feature_type(x) for x in feature_types]
+
+
 class Filter:
     """Specifies filters corresponding to one category to extract from OSM data."""
 
     def __init__(
         self,
-        feature_types: list[FeatureType] | None = None,
+        feature_types: list[FeatureType | str] | None = None,
```
favyen2: Can we avoid passing list[str] here so we can keep it as `list[FeatureType]` and eliminate the need for `_parse_feature_type` and `_coerce_feature_types`? Also `FeatureType | str` doesn't work too well, if it really needs to be union then it may need to be Sequence[FeatureType | str].

rslearn/vis/utils.py
```diff
@@ -100,6 +101,27 @@ def format_window_info(
...
+
+    Args:
+        window: Window object
+
+    Returns:
+        A GeoJSON geometry mapping (e.g. {"type": "Polygon", "coordinates": [...]})
+        with WGS84 lon/lat coordinates, or None on failure.
+    """
+    try:
+        geom_wgs84 = window.get_geometry().to_projection(WGS84_PROJECTION)
```
favyen2: I think you can just inline this since it's just two lines of code (instead of adding this new window_geometry_geojson function). The try/except seems unnecessary since there should not normally be errors building GeoJSON for the window.

rslearn/vis/vis_server.py
```diff
@@ -28,10 +31,52 @@
...
+# talks to this Flask app, and the server fetches the external resources.
+_HTTP_UA = "rslearn-vis-server/1.0 (+https://github.com/allenai/rslearn)"
+_HTTP_TIMEOUT = 15.0
+_HTTP_CACHE_MAX = 8192
+_HTTP_CACHE: OrderedDict[str, tuple[bytes, str]] = OrderedDict()
+_HTTP_CACHE_LOCK = threading.Lock()
+
+# Leaflet assets proxied via /vendor/<name>.
+_VENDOR_SOURCES = {
+    "leaflet.js": "https://unpkg.com/leaflet@1.9.4/dist/leaflet.js",
```
favyen2: Is there an advantage to not needing the browser to have external Internet access? Otherwise it would simplify things to just reference these URLs directly in the visualization.html template.

rslearn/vis/vis_server.py
```diff
@@ -28,10 +31,52 @@
...
+    "leaflet.css": "https://unpkg.com/leaflet@1.9.4/dist/leaflet.css",
+}
+# OpenStreetMap tile template proxied via /basemap/<z>/<x>/<y>.png.
+_TILE_URL_TEMPLATE = "https://{s}.tile.openstreetmap.org/{z}/{x}/{y}.png"
+
+
+def _http_get_cached(url: str) -> tuple[bytes, str] | None:
```
favyen2: Don't need this if the vendor sources are referenced directly.

rslearn/vis/vis_server.py
```diff
@@ -196,6 +243,38 @@ def index() -> str:
...
+        body, _ = result
+        mimetype = "text/css" if name.endswith(".css") else "application/javascript"
+        return Response(
+            body,
+            mimetype=mimetype,
+            headers={"Cache-Control": "public, max-age=86400"},
+        )
+
+    @app.route("/basemap/<int:z>/<int:x>/<int:y>.png")
+    def basemap_tile(z: int, x: int, y: int) -> Response:
```
favyen2: Similarly I would just reference the basemap directly in the template so it doesn't need to be proxied through the visualization server.

tests/integration/data_sources/test_openstreetmap.py
```diff
@@ -105,3 +108,55 @@ def test_ingest(tmp_path: pathlib.Path, test_pbf: pathlib.Path) -> None:
...
+    with pytest.raises(FileNotFoundError, match="missing on disk"):
+        OpenStreetMap(
+            pbf_fnames=["a.osm.pbf", "b.osm.pbf"],
+            bounds_fname="bounds.json",
+            categories={"building": Filter(tag_conditions={"building": []})},
+            context=DataSourceContext(ds_path=UPath(tmp_path)),
+        )
+
+
+def test_openstreetmap_jsonargparse_accepts_osm_feature_type_strings(
```
favyen2: jsonargparse should already parse the strings. It may just need to be all caps ("WAY" instead of "way").

favyen2 (approved):

PR #667, opened 2026-06-10
Avoid intersection-sized temporaries in read_raster_window_from_tiles

rpw1134:
When materializing whole-scene windows (e.g. Sentinel-1 vessel detection, 25,630×20,455 px), `read_raster_window_from_tiles` allocates several multi-GB intersection-sized temporaries on top of the source and destination arrays: the `src[src_indexes]` fancy-index copy, an unconditional `astype()` copy, an `out_crop[dst_indexes]` copy in the mask computation, and per-band masked-gather copies. Materializing a two-item historical layer peaked at **26.7 GiB** anon, OOM-killing our inference container on nodes with ~26 GiB allocatable.

This rewrites the write path to work band-by-band on views: the first-valid mask is computed incrementally per band, and writes go through `np.copyto(..., where=mask, casting="unsafe")`, which converts dtype on the fly with the same semantics as the `astype()` it replaces. Peak extra allocation is now one (T, H, W) bool mask.

Verified on the production scene that triggered the OOM (target + 2 historicals, image-mode predict_pipeline from rslearn_projects, v0.0.23 image with this file patched in): peak anon dropped **26.7 → 20.4 GiB**, identical predictions (114 detections, matching lat/lon/score), ~1 min faster wall clock.

Behavior is unchanged; adds `tests/unit/dataset/test_tile_utils.py` covering first-valid masking (incl. partially-valid pixels and NaN nodata), dtype conversion, band reordering across band sets, partial intersection offsets, remapping, and the `nodata_val=None` path.

### DIFF
--- a/rslearn/dataset/tile_utils.py
+++ b/rslearn/dataset/tile_utils.py
@@ -146,30 +146,35 @@ def read_raster_window_from_tiles(
         dst_col = intersection[0] - bounds[0]
         dst_row = intersection[1] - bounds[1]
 
-        src_sel = src[src_indexes, :, :, :]  # (C_sel, T, H_int, W_int)
-        if remapper:
-            src_sel = remapper(src_sel, band_dtype)
-
         out_crop = dst.array[
             :,
             :,
-            dst_row : dst_row + src_sel.shape[2],
-            dst_col : dst_col + src_sel.shape[3],
+            dst_row : dst_row + src.shape[2],
+            dst_col : dst_col + src.shape[3],
         ]  # (C, T, H_int, W_int) view
 
-        src_typed = src_sel.astype(band_dtype)
-
+        # Work band-by-band on views and write with np.copyto so that no
+        # intersection-sized temporary is ever materialized. For whole-scene
+        # windows the fancy-indexing copies (src[src_indexes], astype, the
+        # masked gathers) each cost multiple GB and dominated peak memory.
         if nodata_val is not None:
             # First-valid: only overwrite pixels where all dst bands are nodata.
-            mask = nodata_eq(out_crop[dst_indexes], nodata_val).min(
-                axis=0
-            )  # (T, H_int, W_int)
-
-            for src_index, dst_index in enumerate(dst_indexes):
-                out_crop[dst_index][mask] = src_typed[src_index][mask]
+            mask = nodata_eq(out_crop[dst_indexes[0]], nodata_val)
+            for dst_index in dst_indexes[1:]:
+                mask &= nodata_eq(out_crop[dst_index], nodata_val)
         else:
-            for src_index, dst_index in enumerate(dst_indexes):
-                out_crop[dst_index] = src_typed[src_index]
+            mask = None
+
+        for src_index, dst_index in zip(src_indexes, dst_indexes, strict=True):
+            src_band = src[src_index]  # (T, H_int, W_int) view
+            if remapper:
+                src_band = remapper(src_band, band_dtype)
+            if mask is not None:
+                # casting="unsafe" matches the astype() conversion semantics
+                # this replaces.
+                np.copyto(out_crop[dst_index], src_band, where=mask, casting="unsafe")
+            else:
+                np.copyto(out_crop[dst_index], src_band, casting="unsafe")
 
     return dst
 
--- a/rslearn/dataset/tile_utils.py
+++ b/rslearn/dataset/tile_utils.py
@@ -153,25 +153,29 @@ def read_raster_window_from_tiles(
             dst_col : dst_col + src.shape[3],
         ]  # (C, T, H_int, W_int) view
 
-        # Work band-by-band on views and write with np.copyto so that no
-        # intersection-sized temporary is ever materialized. For whole-scene
-        # windows the fancy-indexing copies (src[src_indexes], astype, the
-        # masked gathers) each cost multiple GB and dominated peak memory.
+        # Build the first-valid mask: a pixel may be written only where it is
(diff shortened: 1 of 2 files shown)


rslearn/dataset/tile_utils.py
```diff
@@ -146,30 +146,39 @@ def read_raster_window_from_tiles(
...
-        src_sel = src[src_indexes, :, :, :]  # (C_sel, T, H_int, W_int)
-        if remapper:
-            src_sel = remapper(src_sel, band_dtype)
-
         out_crop = dst.array[
             :,
             :,
-            dst_row : dst_row + src_sel.shape[2],
-            dst_col : dst_col + src_sel.shape[3],
+            dst_row : dst_row + src.shape[2],
```
rpw1134, line 152: Three eager whole-stack copies are removed just above this line: `src[src_indexes]` (fancy-index copy of all selected bands), the `remapper(...)` copy, and `src_typed = src_sel.astype(...)`. For a whole-scene window each is multiple GB and they were live at once — that was the peak-memory / OOM driver. This slice now reads H/W from `src.shape` instead of the deleted `src_sel`; equivalent, since band selection never changes H/W. `out_crop` itself is a view into the output, not a copy.

rslearn/dataset/tile_utils.py
```diff
@@ -146,30 +146,39 @@ def read_raster_window_from_tiles(
...
-            mask = nodata_eq(out_crop[dst_indexes], nodata_val).min(
-                axis=0
-            )  # (T, H_int, W_int)
-
-            for src_index, dst_index in enumerate(dst_indexes):
-                out_crop[dst_index][mask] = src_typed[src_index][mask]
+            mask = nodata_eq(out_crop[dst_indexes[0]], nodata_val)
+            for dst_index in dst_indexes[1:]:
+                mask &= nodata_eq(out_crop[dst_index], nodata_val)
         else:
```
rpw1134, line 164: First-valid mask, built incrementally. It marks pixels that are nodata in **every** destination band — only those may be overwritten. Start from band 0's nodata mask, then logically-and (`&=`) each remaining band in: a pixel valid in *any* band becomes False and is left untouched. This replaces the old `nodata_eq(out_crop[dst_indexes], nodata_val).min(axis=0)`, which gathered all bands into a new array before reducing — same boolean result, but no multi-band gather temporary.

rslearn/dataset/tile_utils.py
```diff
@@ -146,30 +146,39 @@ def read_raster_window_from_tiles(
...
-            for src_index, dst_index in enumerate(dst_indexes):
-                out_crop[dst_index] = src_typed[src_index]
+            # nodata_val is None: every source pixel overwrites the destination.
+            mask = None
+
+        # Copy one band at a time. src[src_index] is a view (no copy), and
+        # np.copyto writes straight into the destination window, so no
+        # intersection-sized temporary array is allocated.
+        for src_index, dst_index in zip(src_indexes, dst_indexes, strict=True):
```
rpw1134, line 171: `zip(..., strict=True)` errors if `src_indexes` and `dst_indexes` differ in length — a safety check the previous `enumerate(dst_indexes)` could not give.

rslearn/dataset/tile_utils.py
```diff
@@ -146,30 +146,39 @@ def read_raster_window_from_tiles(
...
+            src_band = src[src_index]  # (T, H_int, W_int) view
```
rpw1134, line 172: Indexing with a single int (`src[src_index]`) returns a **view**, not a copy — this is the crux of the memory fix. The remap, when present, now runs on this one band only (band-sized), not the whole stack.

rslearn/dataset/tile_utils.py
```diff
@@ -146,30 +146,39 @@ def read_raster_window_from_tiles(
...
+            if remapper:
+                src_band = remapper(src_band, band_dtype)
+            if mask is not None:
+                # where=mask restricts the write to first-valid pixels;
+                # casting="unsafe" converts src_band's dtype to band_dtype
+                # during the copy (e.g. uint16 -> float32).
+                np.copyto(out_crop[dst_index], src_band, where=mask, casting="unsafe")
```
rpw1134, line 179: `np.copyto(..., where=mask)` does the masked write **in place** — it replaces the old `out_crop[dst_index][mask] = src_typed[src_index][mask]`, which allocated temporaries on both sides (masked gather + masked scatter). `casting="unsafe"` performs the dtype conversion during the copy (e.g. uint16 -> float32), reproducing the exact semantics of the deleted `.astype(band_dtype)`.

tests/unit/dataset/test_tile_utils.py
```diff
@@ -0,0 +1,226 @@
...
+        name=name,
+        geometry=STGeometry(
+            projection=PROJECTION,
+            shp=Polygon([(0, 0), (10, 0), (10, 10), (0, 10)]),
+            time_range=None,
+        ),
+    )
+
+
+def write(
```
favyen2: These helpers don't seem to be saving very many arguments. So I would just inline it.

favyen2 (approved):
looks good, thanks for the memory optimization!

tests/unit/dataset/test_tile_utils.py
rpw1134: Done — inlined the write helper into tile_store.write_raster(...) at the call sites in af68571. Left read in place since it also wraps TileStoreWithLayer and fills in bounds plus several kwargs, so it cuts more boilerplate; happy to inline it too if you would prefer consistency.

PR #669, opened 2026-06-15
[New feature] Add GeoParquet support for Sentinel2 from PC

robmarkcole:
Adds a GeoParquet metadata backend to the Planetary Computer data sources. Previously, scene discovery during dataset prepare always used the Planetary Computer STAC API, making one search request per window, which can hit API rate limits on large jobs.

```
Retrying after catching error in retry loop: 503 Server Error: Operations per second is over the account limit.
```

The new backend queries the collection-level GeoParquet item table (hosted on Azure Blob Storage) using DuckDB. It downloads only the relevant date-range partitions, runs a spatial and temporal filter locally, and returns matching items in a single bulk operation per `get_items`call. Because rslearn already batches all windows into one `get_items` call during prepare, this effectively replaces N per-window STAC requests with one bulk query.

The feature is opt-in via a new metadata_backend parameter (default "stac", new option "geoparquet") on the PlanetaryComputer base class and all its subclasses including Sentinel2. An optional `metadata_cache_dir` can be set to cache the partition file list and query results between runs. DuckDB is required and is included in the existing `extra` optional dependencies.

- Implemented functions to create mock GeoParquet rows for Sentinel2.
- Added tests for batch preparation of window metadata and item retrieval by name.
- Updated `uv.lock` to include duckdb version 1.5.3 and incremented revision number.
- Added duckdb as an extra dependency for the project.

**Note:** I need to apply this to a large-scale dataset to see if there is a significant speedup

### DIFF
--- a/docs/data_sources/planetary_computer_Sentinel2.md
+++ b/docs/data_sources/planetary_computer_Sentinel2.md
@@ -18,12 +18,46 @@ Sentinel-2 satellite images from https://planetarycomputer.microsoft.com/dataset
     "query": null,
     "sort_by": null,
     "sort_ascending": true,
+    // Metadata backend for scene discovery. The default "stac" uses the Planetary
+    // Computer STAC API. Set "geoparquet" to query the collection's GeoParquet item
+    // table instead, which avoids per-window STAC search requests during prepare.
+    "metadata_backend": "stac",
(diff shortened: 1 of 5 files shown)


robmarkcole:
not doing

PR #674, opened 2026-06-23
Read raster composites in spatial blocks to bound materialize memory

rpw1134:
## Motivation

`read_raster_window_from_tiles` reads each item's whole intersection in one `tile_store.read_raster` call, allocating a full-scene source array on top of the full-scene `dst` accumulator. On large scenes (~25000×20000 Sentinel-1, 2 bands f32) that transient dominates peak memory and OOM-kills materialize. This reads/composites the intersection in spatial blocks so the source transient is bounded to ~one block.

## Change
- New `TileStore.read_raster_blocked` generator. Base default yields the whole bounds as a single block — non-overriding (e.g. remote/API-backed) stores keep their exact one-read behavior. `DefaultTileStore` overrides it to open the COG + `WarpedVRT` once and read each block as a window of that shared VRT (`GeotiffRasterFormat.decode_raster_blocked`), avoiding per-block reopen.
- `read_raster_window_from_tiles` now loops over `read_raster_blocked`; composite math unchanged.

## Correctness
Bit-identical to the single-read path (the warp is per-output-pixel, independent of read window). Covered by new parametrized multi-block regression tests (nearest/bilinear/cubic, T>1, nonzero offsets, first-valid nodata) and verified end-to-end on a Sentinel-1 vessel scene (114/114 detections unchanged).

## Measured (Sentinel-1, ~25k×20k, 2 historicals)
Materialize anon peak **~26.7 → ~14–16 GiB**. Blocked reads add ~one extra knob: the shared full-scene `WarpedVRT` peaks ~15.96 GiB and is ~12% faster on materialize; per-block VRTs peak ~14.46 GiB and are slower. Both are well under the prior single-read peak.

### DIFF
--- a/rslearn/dataset/tile_utils.py
+++ b/rslearn/dataset/tile_utils.py
@@ -18,6 +18,13 @@ from .remap import Remapper
 if TYPE_CHECKING:
     from rslearn.data_sources.data_source import ItemType
 
+# Spatial block size for windowed reads in read_raster_window_from_tiles. Reading a
+# whole-scene intersection at once allocates a full-scene source array plus rasterio
+# warp buffers, which dominate peak memory on large scenes; reading in blocks bounds
+# the transient warp allocation to roughly one block. 2048 keeps that transient small;
+# larger blocks raise the per-block warp buffer with no offsetting benefit.
+_READ_BLOCK_SIZE = 2048
+
 
 def get_needed_band_sets_and_indexes(
     item: ItemType,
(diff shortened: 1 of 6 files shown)


favyen2:
This approach looks reasonable to me -- only add blocked reading to `GeotiffRasterFormat` since it can be used directly in `DefaultTileStore`, and the default implementation in `TileStore` reuses `read_raster` (one read). I think that design is good.

I still like the alternative of having the downstream user create smaller windows though. Since it feels like we have converged on the window being the unit of parallelization/processing. I think the recall for Sentinel-1 vessel detection was degraded but if that's an issue then it seems like overlapping windows would solve it, although I thought that would be unnecessarily complex; 5% drop is higher than what I expected though.

Also one limitation is it's really only reducing memory usage by up to 1/2 ish because `dst` is still the size of the window.

I'm curious what others think, @APatrickJ do you have any thoughts about it? Otherwise I think it's fine as long as it is validated that there isn't a slowdown for windows smaller than the block size, e.g. when materializing thousands of 64x64 windows.

rslearn/dataset/tile_utils.py
```diff
@@ -18,6 +18,13 @@
 if TYPE_CHECKING:
     from rslearn.data_sources.data_source import ItemType
 
+# Spatial block size for windowed reads in read_raster_window_from_tiles. Reading a
+# whole-scene intersection at once allocates a full-scene source array plus rasterio
+# warp buffers, which dominate peak memory on large scenes; reading in blocks bounds
+# the transient warp allocation to roughly one block. 2048 keeps that transient small;
+# larger blocks raise the per-block warp buffer with no offsetting benefit.
+_READ_BLOCK_SIZE = 2048
```
APatrickJ, line 26: This seems to be duplicated from `raster_format.DEFAULT_READ_BLOCK_SIZE`

favyen2:
We decided to use stopgap solution instead, where we split up vv and vh into separate band sets so they are materialized into two separate GeoTIFFs thereby reducing memory consumption. More permanent solution is to update vessel detection pipelines to use smaller windows instead of just one window for the whole scene.

PR #574, opened 2026-03-30
Add lat/lon support to OlmoEarth finetuning wrapper

pjreddie:
## Summary
- Adds `use_latlon: bool` parameter to `OlmoEarth` wrapper
- When enabled, computes lat/lon from `SampleMetadata` crop center + CRS projection via `rasterio.warp.transform`
- Sets `latlon` on `MaskedOlmoEarthSample` so the encoder applies geographic encoding during finetuning
- Requires pretrained model trained with `use_latlon_encoding=True`

## Related PRs
- allenai/olmoearth_pretrain#518 — lat/lon encoding in pretraining
- gea: latlon experiment config for ecosystem segmentation (forthcoming)

## Test plan
- [x] Unit tests: WGS84 identity transform, UTM non-trivial transform, batch processing
- [x] Integration tests: full forward pass with `use_latlon=True`, backward compat with `use_latlon=False`
- [x] All 17 existing OlmoEarth tests still pass

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #677, opened 2026-06-26
Add use_latlon support to OlmoEarth fine-tuning wrapper

pjreddie:
## Summary
Adds a `use_latlon` option to the `OlmoEarth` fine-tuning wrapper. When enabled, each sample's crop-center lat/lon is computed from its `SampleMetadata` (`crop_bounds` + `projection` CRS, transformed to WGS84) and passed to the encoder via `MaskedOlmoEarthSample.latlon`, so models pretrained with a geographic (lat/lon) encoding (e.g. the `separate`/simple-encoding OlmoEarth variants) can use it during fine-tuning.

The latlon encoding's dropout is train-only, and some checkpoints were pretrained with a non-zero `latlon_dropout_rate` (e.g. `rope_simple_v1` used 0.5). To keep the encoding active on every fine-tuning step, the wrapper forces `latlon_dropout_rate = 0` when `use_latlon` is set.

`use_latlon` defaults to `False`, so existing configs are unaffected.

## Changes
- `OlmoEarth.__init__`: new `use_latlon: bool = False` arg; when set (and the loaded encoder exposes `composite_encodings.latlon_dropout_rate`), force the rate to 0.
- New `OlmoEarth._compute_latlon_from_metadata(...)`: crop-center → WGS84 `(lat, lon)` per sample.
- `OlmoEarth.forward`: when `use_latlon`, inject the computed lat/lon into the sample before the encoder.

## Testing
- Loaded a lat/lon-pretrained checkpoint (`rope_simple_v1`) with `use_latlon=True`; confirmed `latlon_dropout_rate` is forced to 0.
- `fast_dev_run` of a segmentation fine-tune (encoder + `UNetDecoder` + `SegmentationHead`) with `use_latlon: true`, `use_legacy_timestamps: false`, and `UNetDecoder(use_batch_norm: true)` completes a train+val step (loss ≈ ln(num_classes) at init, as expected).

### DIFF
--- a/rslearn/models/olmoearth_pretrain/model.py
+++ b/rslearn/models/olmoearth_pretrain/model.py
@@ -8,6 +8,7 @@ from dataclasses import dataclass
 from datetime import datetime, timedelta
 from typing import Any
 
+import shapely
 import torch
 from einops import rearrange
 from olmoearth_pretrain_minimal import ModelID, load_model_from_id, load_model_from_path
@@ -25,10 +26,12 @@ from olmoearth_pretrain_minimal.olmoearth_pretrain_v1.utils.datatypes import (
 )
 from upath import UPath
 
+from rslearn.const import WGS84_PROJECTION
 from rslearn.log_utils import get_logger
 from rslearn.models.component import FeatureExtractor, FeatureMaps, TokenFeatureMaps
 from rslearn.models.olmoearth_pretrain.norm import OlmoEarthNormalize
 from rslearn.train.model_context import ModelContext, RasterImage, SampleMetadata
+from rslearn.utils.geometry import STGeometry
 
 logger = get_logger(__name__)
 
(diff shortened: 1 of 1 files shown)


rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -226,6 +243,38 @@ def _patch_legacy_encoder_config(self, config_dict: dict) -> dict:
...
+        (EPSG:4326) to produce geographic coordinates for the lat/lon encoding.
+
+        Args:
+            metadatas: list of SampleMetadata, one per sample in the batch.
+            device: torch device to place the output tensor on.
+
+        Returns:
+            Tensor of shape (B, 2) with (latitude, longitude) in degrees.
+        """
+        from rasterio.crs import CRS
```
favyen2: Please see the comments on #574, they apply here as well (move imports to top of file and use STGeometry for the re-projection).

rslearn/models/olmoearth_pretrain/model.py
```diff
@@ -213,6 +219,17 @@ def __init__(
...
+        self.use_latlon = use_latlon
+        # During fine-tuning we want the geographic (lat/lon) encoding active on every
+        # step. Models trained with a latlon encoding may carry a non-zero
+        # latlon_dropout_rate (e.g. rope_simple_v1 used 0.5); dropout is train-only, so
+        # force it to 0 so lat/lon is ingested on every training step rather than being
+        # randomly zeroed.
+        if use_latlon and hasattr(self.model, "composite_encodings"):
+            encodings = self.model.composite_encodings
+            if hasattr(encodings, "latlon_dropout_rate"):
+                encodings.latlon_dropout_rate = 0.0
```
favyen2: We changed `band_dropout_rate` to be the reverse, where it is initially set to 0.0 and only set to the configured value during training. Can we do the same here, so that users of olmoearth_pretrain don't need to worry about disabling `latlon_dropout_rate`?

favyen2:
Let me know when this is ready for review, currently it is marked draft

PR #679, opened 2026-07-10
Add mkdocs website allenai.github.io/rslearn

favyen2:
This adds an initial mkdocs website.

Next step is to add more content to the Usage tab, currently it is pretty bare. I want to structure Usage like this:

- Usage
- Specifying the Dataset Config
- Create Windows
  - Create Windows via CLI
  - Create Windows Programmatically
- Add Raster and Vector Data Manually
  - Raster Data
  - Vector Data
- Specifying the Model Config

So each section would have some quick basic general usage, not tied to a specific example.

Specifying the X Config can mention both the examples and the Config Reference, e.g. Model Config can mention example of simple case in the Quickstart example while also mentioning we have docs on each foundation model that includes per-task model config examples that could be borrowed from.

Anyway that would be in future PR, this one just creates the initial doc website with a few doc changes/improvements.

See `Development.md` about building and serving documentation locally to test it:

```bash
pip install .[docs]
mkdocs serve
```


docs/examples/Quickstart.md
gabrieltseng, line 118: nit: if we could add a screenshot of the tif here it would break the text up a bit and make it easier to follow

docs/examples/Quickstart.md
```diff
@@ -1,19 +1,9 @@
-## Introductory Example
+## Quickstart
```
gabrieltseng, line 1: I actually think the stadiums example is a better "quickstart" - having dense labels like WorldCereal strikes me as relatively unusual so this seems less "start and go" than the stadiums example.

docs/examples/Quickstart.md
```diff
@@ -187,81 +190,103 @@ model:
...
+    # features back to the input resolution, and a segmentation head that predicts a
+    # land cover class for each pixel.
     model:
       class_path: rslearn.models.singletask.SingleTaskModel
       init_args:
         encoder:
-          - class_path: rslearn.models.satlaspretrain.SatlasPretrain
+          # This applies the OlmoEarth encoder on the inputs. We use the v1.2 Nano
+          # model here since it is fast; larger models generally give better accuracy.
+          # Switch to TINY or BASE for better results.
```
gabrieltseng: nit: "Switch to TINY, SMALL or BASE..."

docs/examples/WindowsFromGeojson.md
```diff
@@ -9,13 +9,6 @@ should be able to substitute any GeoJSON (although model performance will vary):
 
 https://pub-956f3eb0f5974f37b9228e0a62f449bf.r2.dev/outputs/marine/latest.geojson
```
gabrieltseng: this url doesn't work for me:

Image

gabrieltseng (approved):

gabrieltseng:
thanks Favyen - this looks great. If you'd prefer to merge this in as is and save those changes for later thats fine too.

docs/examples/Quickstart.md
favyen2: I agree that the stadiums example would be a better fit for some users, but there are many ways to use rslearn so currently the main README suggests looking through the examples and finding one most similar to the task they want to do. For the quickstart, I think the current one is better because it is quicker with no Python code needed to walk through it, while still giving an intro to concepts like windows and how prepare/materialize match windows against data source items.

docs/examples/WindowsFromGeojson.md
favyen2: fixed, thanks!

docs/examples/Quickstart.md
favyen2: updated!

docs/examples/Quickstart.md
favyen2: added images a bit lower under the Adding Land Cover Labels section

PR #681, opened 2026-07-20
Re-structure docs so we have simple quick examples in Usage

favyen2:
- Modernize all the dataset and model configs, e.g. remove deprecated `cache_dir` arg for Planetary Computer data sources, use `num_timesteps_per_forward_pass` instead of `image_channels` with `SimpleTimeSeries`, use `nodata_value` instead of `nodata_vals`, etc.
- Limit model/dataset config comments to 80 characters so it fits in the narrower default mkdocs view.
- Add quick usage examples: `CreateWindowsProgrammatically.md`, `AddRasterData.md`, and `AddVectorData.md`.
- Refactor the `ProgrammaticWindows.md` tutorial since the first part of it became `CreateWindowsProgrammatically.md`. New the second part is the `ConvertEuroSAT.md` example.


docs/examples/NaipSentinel2.md
APatrickJ: We should update this to use the `storage` parameter

docs/examples/BitemporalSentinel2.md
APatrickJ: `open_layer_writer` doesn't exist on `Window`. Maybe this should be:

```python
with window.data.open_layer_writer("label") as writer:
```

docs/ModelConfig.md
```diff
@@ -539,42 +549,44 @@ Here is a summary of all of the options available in the SplitConfig.
...
-      # can be cropped using transforms, but if a random crop is desired, it is more
-      # efficient to crop it with this option, since this way the cropping will happen
-      # when reading GeoTIFFs. However, setting it here is less flexible, since it only
-      # supports random cropping.
+      # By default (patch_size=null), data for the entire window bounds is read.
+      # This can be cropped using transforms, but if a random crop is desired,
+      # it is more efficient to crop it with this option, since this way the
+      # cropping will happen when reading GeoTIFFs. However, setting it here is
+      # less flexible, since it only supports random cropping.
       patch_size: 128
```
APatrickJ: Should this be `crop_size` now that `patch_size` is deprecated?

mkdocs.yml
```diff
@@ -79,20 +79,26 @@ markdown_extensions:
 # These helper README files sit alongside example data assets and are not pages.
 exclude_docs: |
   examples/BitemporalSentinel2/README.md
+  examples/ProgrammaticWindows.md
```
APatrickJ, line 84: If you want to keep this stub so old links still work, then this should not be excluded as `exclude_docs` drops it from the build entirely

Or we can make it an actual redirect if you'd like:

```yaml
plugins:
  - redirects:
      redirect_maps:
        examples/ProgrammaticWindows.md: CreateWindowsProgrammatically.md
```

docs/AddVectorData.md
```diff
@@ -0,0 +1,160 @@
...
+from rslearn.utils.feature import Feature
+from rslearn.utils.geometry import STGeometry
+from rslearn.utils.get_utm_ups_crs import get_utm_ups_projection
+from rslearn.utils.vector_format import GeojsonVectorFormat
+
+WINDOW_SIZE = 256
+dataset = Dataset(UPath("./dataset"))
+
+with open("points.csv") as f:
+    rows = list(csv.DictReader(f))
```
APatrickJ: The rest of this example uses 2 space indent, might want to make that consistent here

docs/AddRasterData.md
```diff
@@ -0,0 +1,172 @@
...
+  # a box in integer pixel coordinates.
+  projection: Projection
+  bounds: tuple[int, int, int, int]
+  projection, bounds = get_raster_projection_and_bounds(source)
+
+  one_hot = source.read()
+
+# Convert the one-hot array into class IDs that are compatible with
+# SegmentationTask. We use 255 as a nodata value.
+valid = np.any(one_hot != 0, axis=0)
```
APatrickJ, line 119: We might want to explain the choice of 0 here, maybe something like "pixels with no active band are treated as nodata"

docs/AddRasterData.md
```diff
@@ -0,0 +1,172 @@
...
+rslearn dataset ingest --root ./dataset --enabled-layers image
+rslearn dataset materialize --root ./dataset --enabled-layers image
+```
+
+The resulting rasters are aligned to each window's projection and bounds.
+
+## Write Rasters Programmatically
+
+Write rasters programmatically when more customizability is needed. For example,
+suppose `one_hot_labels.tif` contains one band per class. `SegmentationTask`
```
APatrickJ, line 57: I might say "suppose `one_hot_labels.tif` contains one band per class (here, 4 classes)." to explain the choice of `num_classes: 4` below

APatrickJ (approved):

mkdocs.yml
favyen2: I think I need it to redirect on github link (https://github.com/allenai/rslearn/blob/master/docs/examples/ProgrammaticWindows.md), I think the redirect_maps won't do that. I think it can be excluded from the build since https://allenai.github.io/rslearn/examples/ProgrammaticWindows/ has only been up for a few days so it is unlikely that removing it now will break existing links. I updated the comment to mention the URL that it is redirecting from.

docs/examples/NaipSentinel2.md
favyen2: thanks, fixed!

docs/examples/BitemporalSentinel2.md
favyen2: thanks, fixed!

docs/ModelConfig.md
favyen2: fixed

docs/AddVectorData.md
favyen2: updated indent across the docs

docs/AddRasterData.md
favyen2: added a comment

docs/AddRasterData.md
favyen2: added a comment above the `num_classes: 4`

Issue #536, opened 2026-02-27
[Discussion] Support for loggers other than wandb


robmarkcole:
I've briefly investigated supporting MLFlow as a self hosted alternative to wandb (which requires a license for commercial usage). The current approach couples “logging” to the wandb API and to wandb-specific assumptions, which makes it hard to add other loggers cleanly.

- Non-scalar metric logging is implemented via `wandb.plot` e.g., confusion matrices. Other loggers (MLflow, TensorBoard, CSV) don’t necessarily have the same plotting API, so you need per-backend implementations or a generic artifact representation.
- The “is wandb initialized?” checks encode a wandb-specific lifecycle. Other backends have different notions of an active run, run IDs, and resume semantics.
- Project management/resume currently stores a `wandb_id` and tries to resume by setting `logger.init_args.id`. That only makes sense for WandbLogger-like backends; MLflow uses `run_id`, others have no equivalent.
- Automatically injecting/configuring a specific logger inside the CLI (rather than letting Lightning instantiate the logger from config) creates a second “source of truth” for logging behavior, which becomes a combinatorial mess once you support multiple backends.

If the stance is to only support wandb in rslearn, I appreciate how that keeps things simple but it would be good to clarify this.

PR #682, opened 2026-07-21
Suport mlflow logger

robmarkcole:
Address https://github.com/allenai/rslearn/issues/536

Validated agains an local instance run with 
```
uv run python -m mlflow server \
  --host 0.0.0.0 \
  --port 8080 \
  --allowed-hosts '*' \
  --cors-allowed-origins '*'
```
image

image

I prototyped logging the confusion matrix as a figure but removed as this adds matplotlib as a dependency. Can add back if of interest

### DIFF
--- a/docs/ModelConfig.md
+++ b/docs/ModelConfig.md
@@ -708,9 +708,9 @@ If enabled, model management will:
    This is used by `ManagedBestLastCheckpoint` to resolve its checkpoint directory.
 2. If training is restarted, resume from the last checkpoint.
 3. During test/predict, automatically load the best checkpoint.
-4. Enable W&B logging and save the W&B run ID to the save project directory (so it can
-   be reused when resuming training).
-5. Save the model config with the W&B run.
+4. Enable W&B or MLflow logging and save the run ID to the project directory (so it
+   can be reused when resuming training).
+5. Save the model config with the experiment run.
 
 To save checkpoints, add a `ManagedBestLastCheckpoint` callback to `trainer.callbacks`.
 This callback automatically determines its checkpoint directory from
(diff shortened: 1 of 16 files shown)


docs/ModelConfig.md
```diff
@@ -755,6 +755,21 @@ trainer:
...
+Lightning's MLflow logger:
+
+```yaml
+trainer:
+  logger:
+    class_path: lightning.pytorch.loggers.MLFlowLogger
+    init_args: {}
+```
+
+Model management fills in `experiment_name` and `run_name`, records the run ID for
```
favyen2: I think this paragraph could be removed since it is mostly re-iterating the 1/2/3/4/5 above but that upper list is more general-purpose since it now covers both W&B and MLFlow.

rslearn/train/metrics.py
```diff
@@ -29,6 +30,16 @@ def log_to_wandb(self, name: str) -> None:
         """
         pass
 
+    def log_to_mlflow(self, name: str, client: Any, run_id: str) -> None:
```
favyen2, line 31: should both of these be abstract method where by default they raise NotImplementedError? or at least the two should be symmetric

favyen2 (commented):
This is great, thanks for adding mlflow support! just had two comments.

favyen2:

rslearn/train/metrics.py
robmarkcole: Made both hooks symmetric:
- Both now default to logging an “unsupported” warning.
- Removed unnecessary ABC/abstractmethod.
- Added a test covering both defaults.

robmarkcole:
@favyen2 I've used this in anger now and addressed a bug

rslearn/lightning_cli.py
```diff
@@ -23,10 +29,62 @@
 from rslearn.utils.jsonargparse import init_jsonargparse
 
 WANDB_ID_FNAME = "wandb_id"
+MLFLOW_ID_FNAME = "mlflow_id"
 
 logger = get_logger(__name__)
 
 
+class RslearnSaveConfigCallback(SaveConfigCallback):
```
favyen2: Is this to fix the config saving to the logger's default save_dir which we are not overriding (e.g. `./config.yaml` when using WandbLogger)? Should we set the save_dir to match the project management dir (configured under default_root_dir)? And if we do that, can we use the default `SaveConfigCallback` or do we still need a custom class? I guess we might still need a custom class in case the project management dir is not on the local filesystem?

Maybe we can add the MLFlow stuff in this PR and figure out the save config callback fix in another PR?

robmarkcole: Agreed, from this branch I now removed callback commits, and added a temporary remote-MLflow config-saving disable

favyen2 (approved):
Thanks for adding MLFlow support, and for iterating on it, this looks good to me now!

PR #685, opened 2026-07-23
Increase ingest speed and memory utilization when only some layers are enabled

favyen2:
Ingest is optimized by only keeping the WindowLayerDatas that match the enabled layers. This improves not only memory utilization, but also can improve speed significantly, probably because it doesn't have to spend time continuously allocating more memory.

I also included a jsonargparse fix in case the ResolutionFactor is configured via class_path with no init_args set.


yawenzzzz (approved):
LGTM!

PR #686, opened 2026-07-23
Atomic reads for AEF csv

gabrieltseng:
Previously, if I kicked off AEF downloads with multiple workers, then they'd all try to download the AEF csv (or check if it exists), and then try to read it. If it hadn't downloaded yet I would get this error:

```
Traceback (most recent call last):
  File "/weka/dfive-default/gabrielt/helios/scripts/tools/materialize_aef_supplemental_embeddings.py", line 127, in <module>
    main()
  File "/weka/dfive-default/gabrielt/helios/scripts/tools/materialize_aef_supplemental_embeddings.py", line 104, in main
    manifest = materialize_product(
               ^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/olmoearth_pretrain/evals/embedding_materializer/materialize.py", line 165, in materialize_product
    for status, window_id in outcomes:
                             ^^^^^^^^
  File "/opt/miniconda3/lib/python3.12/concurrent/futures/_base.py", line 619, in result_iterator
    yield _result_or_cancel(fs.pop())
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda3/lib/python3.12/concurrent/futures/_base.py", line 317, in _result_or_cancel
    return fut.result(timeout)
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda3/lib/python3.12/concurrent/futures/_base.py", line 449, in result
    return self.__get_result()
           ^^^^^^^^^^^^^^^^^^^
  File "/opt/miniconda3/lib/python3.12/concurrent/futures/_base.py", line 401, in __get_result
    raise self._exception
  File "/opt/miniconda3/lib/python3.12/concurrent/futures/thread.py", line 59, in run
    result = self.fn(*self.args, **self.kwargs)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/olmoearth_pretrain/evals/embedding_materializer/materialize.py", line 158, in run_one
    _process_window(window, fetcher, provider, year, overwrite),
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/olmoearth_pretrain/evals/embedding_materializer/materialize.py", line 95, in _process_window
    array = fetcher.fetch(window.bounds, window.projection, target_year)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/olmoearth_pretrain/evals/embedding_materializer/fetchers.py", line 242, in fetch
    groups = source.get_items([geometry], query_config)[0]
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/rslearn/rslearn/data_sources/aws_google_satellite_embedding_v1.py", line 211, in get_items
    grid_index, _ = self._load_index()
                    ^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/rslearn/rslearn/data_sources/aws_google_satellite_embedding_v1.py", line 174, in _load_index
    df = self._read_index_csv()
         ^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/rslearn/rslearn/data_sources/aws_google_satellite_embedding_v1.py", line 161, in _read_index_csv
    return pd.read_csv(cache_file)
           ^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/.venv/lib/python3.12/site-packages/pandas/io/parsers/readers.py", line 1026, in read_csv
    return _read(filepath_or_buffer, kwds)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/.venv/lib/python3.12/site-packages/pandas/io/parsers/readers.py", line 620, in _read
    parser = TextFileReader(filepath_or_buffer, **kwds)
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/.venv/lib/python3.12/site-packages/pandas/io/parsers/readers.py", line 1620, in __init__
    self._engine = self._make_engine(f, self.engine)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/.venv/lib/python3.12/site-packages/pandas/io/parsers/readers.py", line 1898, in _make_engine
    return mapping[engine](f, **self.options)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/weka/dfive-default/gabrielt/helios/.venv/lib/python3.12/site-packages/pandas/io/parsers/c_parser_wrapper.py", line 93, in __init__
    self._reader = parsers.TextReader(src, **kwds)
                   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "pandas/_libs/parsers.pyx", line 581, in pandas._libs.parsers.TextReader.__cinit__
pandas.errors.EmptyDataError: No columns to parse from file
```

This fixes that using an atomic write, which is locked to avoid multiple concurrent downloads.

### DIFF
--- a/rslearn/data_sources/aws_google_satellite_embedding_v1.py
+++ b/rslearn/data_sources/aws_google_satellite_embedding_v1.py
@@ -2,6 +2,7 @@
 
 import os
 import tempfile
+import threading
 from collections.abc import Callable
 from datetime import UTC, datetime
 from typing import Any
@@ -31,11 +32,14 @@ from rslearn.data_sources.direct_materialize_data_source import (
     DirectMaterializeDataSource,
 )
 from rslearn.data_sources.utils import MatchedItemGroup
-from rslearn.utils.fsspec import join_upath
+from rslearn.log_utils import get_logger
+from rslearn.utils.fsspec import join_upath, open_atomic
 from rslearn.utils.geometry import PixelBounds, Projection, STGeometry
 from rslearn.utils.grid_index import GridIndex
 from rslearn.utils.raster_array import RasterArray, RasterMetadata
 from rslearn.utils.raster_format import get_raster_projection_and_bounds
+
+logger = get_logger(__name__)
 
 # Band names for the 64 embedding channels
 BANDS = [f"A{idx:02d}" for idx in range(64)]
@@ -141,9 +145,36 @@ class GoogleSatelliteEmbeddingV1(
             region_name="us-west-2",
         )
 
-        # Lazy-loaded grid index
+        # Lazy-loaded grid index, guarded by a lock so that concurrent readers
+        # (e.g. threads in a materialization pool) download and parse the index
+        # only once.
+        self._index_lock = threading.Lock()
         self._grid_index: GridIndex | None = None
         self._items_by_name: dict[str, GoogleSatelliteEmbeddingV1Item] | None = None
+
+    def _download_index_csv(self, cache_file: UPath) -> None:
+        """Download the index CSV from S3 to the cache file.
+
+        The download is streamed via open_atomic so a killed or concurrent run
+        cannot leave a partial file at the final path, and the resulting size
+        is validated against the S3 object size (deleting the file on
+        mismatch) to catch truncated response bodies.
+
+        Args:
+            cache_file: the final path to place the index CSV at.
+        """
+        response = self.s3_client.get_object(Bucket=BUCKET_NAME, Key=INDEX_KEY)
+        expected_size = response["ContentLength"]
+        with open_atomic(cache_file, "wb") as f:
+            for chunk in response["Body"].iter_chunks(chunk_size=8 * 1024 * 1024):
+                f.write(chunk)
+        actual_size = cache_file.stat().st_size
+        if actual_size != expected_size:
+            cache_file.unlink()
+            raise OSError(
+                f"AEF index download was truncated: got {actual_size} bytes, "
+                f"expected {expected_size}"
+            )
 
     def _read_index_csv(self) -> pd.DataFrame:
         """Read the index CSV, downloading from S3 if not cached.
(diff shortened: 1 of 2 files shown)


favyen2:
We use multiprocessing for rslearn worker parallelism so I think the thread lock might not help.

Passing `--no-use-initial-job` can help for these situations where the data source needs to do some init work: the default behavior (if you don't override it by passing `--no-use-initial-job`) is to process one window first before starting many processes. I think in this case it would avoid the issue with AEF that you ran into. You can simply drop the `--no-use-initial-job` argument and see if that helps.

favyen2:
BTW assuming the `--no-use-initial-job` does help, currently that whole system is super convoluted (how to know which data sources do initial setup that depends on processing one window first before parallelizing, and which data sources are fine with parallelizing from the getgo), I was thinking in the past that we could separate out that init stuff into a setup function and probably get rid of the initial job stuff altogether, never got a chance to do that but if you're interested in taking a look I think it would make it a lot simpler. I think right now the documentation tries to encourage avoiding `--no-use-initial-job` (so always do the one window processing) but then it's slower than it needs to be especially in cases where a single window takes several minutes to materialize.

gabrieltseng:
I wasn't running this using `materialize` (in which case I would have benefited from the `no-use-initial-job` - I was just calling the `GoogleSatelliteEmbeddingV1` directly in a script (the reason for this was to have a script which could download GSE and Tessera directly). 

But yea I think the correct move is to use materialize; I'll close this. Thanks @favyen2

PR #701, opened 2026-08-17
Enhancement: Allow SegmentationTask to use optional class names when logging per-class metrics

### LINKED ISSUE #700: [RF] Add support for semantic class labels in segmentation metric names.
Extend SegmentationTask’s existing optional class_names parameter to label all per-class metrics

robmarkcole:
Closes https://github.com/allenai/rslearn/issues/700

Allow SegmentationTask to use optional class names when logging per-class metrics, following classification’s class-first naming convention (for example, test_wheat_f1). Preserve existing cls_N names when class names are omitted.


rslearn/train/tasks/segmentation.py
```diff
@@ -451,6 +480,8 @@ def __init__(
         pass_probabilities: bool = True,
         class_idx: int | None = None,
         output_key: str | None = None,
+        class_names: list[str] | None = None,
+        class_metric_name: str | None = None,
```
favyen2, line 484: If we exclude this then does the metric name get hidden because we are returning a dict from strings to values? Or can we just rely on that metric name prefix instead of also adding a suffix.

robmarkcole: Here’s the observed TorchMetrics behaviour:

```python
single = MetricCollection({
    "ClasswiseF1": DictMetric({"wheat": 0.8}),
}, prefix="test_")

# {"test_wheat": ...}
# Outer name "ClasswiseF1" is hidden.
```

When multiple metrics return identical dictionary keys:

```python
metrics = MetricCollection({
    "ClasswiseF1": DictMetric({"wheat": 0.8}),
    "ClasswisePrecision": DictMetric({"wheat": 0.9}),
}, prefix="test_")

# {
#   "test_ClasswiseF1_wheat": ...,
#   "test_ClasswisePrecision_wheat": ...,
# }
```

With explicit unique suffixes:

```python
metrics = MetricCollection({
    "ClasswiseF1":
        DictMetric({"wheat_classwise_f1": 0.8}),
    "ClasswisePrecision":
        DictMetric({"wheat_classwise_precision": 0.9}),
}, prefix="test_")

# {
#   "test_wheat_classwise_f1": ...,
#   "test_wheat_classwise_precision": ...,
# }
```

So `class_metric_name` ensures stable names regardless of which metrics are included. The current implementation passes `class_metric_name="classwise_f1"` and produces stable keys such as: `test_wheat_classwise_f1`
Removing class_metric_name would introduce the collision-dependent behavior

favyen2 (approved):

PR #706, opened 2026-09-07
[Bugfix] Make WorldCereal downloads resumable and checksum-verified

### LINKED ISSUE #705: [Bug] WorldCereal data source: 10s download timeout, no retries, and fake "resume support"
**Location:** rslearn/data_sources/worldcereal.py (~lines 54, 381–389), using open_atomic() in rslearn/utils/fsspec.py.

**Problem:**
Downloading WorldCereal archives (multi-GB files from Zenodo) uses a hardcoded 10s `requests.get(..., timeout=10)` with no retry/backoff, so any brief stall aborts the download with ReadTimeout. A comment claims "resume support," but there's no actual resume logic — no Range requests, no partial-byte tracking; it only skips download if the file is already fully complete. Since the write goes through open_atomic() (temp file + os.rename on success), any interrupted download is never renamed into place, so the next attempt restarts from byte 0 — and the orphaned temp file is left on disk uncollected.

**Impact:** rslearn prepare for WorldCereal can reliably fail on slow/unstable connections, can't resume, and leaks temp files with each failed retry.

**Suggested fix:** make the timeout larger/configurable, wrap the request with the repo's existing create_retry_session() helper, implement real Range-based resume (or fix the comment), and clean up orphaned temp files on failure.

robmarkcole:
Closes https://github.com/allenai/rslearn/issues/705

- Split the timeout into CONNECT_TIMEOUT_SECONDS=10 / READ_TIMEOUT_SECONDS=60, and route the request through the repo's existing create_retry_session() helper.
- Added _download_with_resume(), which writes to a stably-named .partial temp file, resumes from the last byte offset via an HTTP Range request on retry (falling back to a full restart if the server doesn't honor Range), and retries transient network errors up to 5 times with backoff.
- Always chunk: _download_with_resume fetches every file as a sequence of bounded Range requests (128 MiB each, RANGE_SIZE_BYTES), even the very first byte of a brand-new download — not just when resuming. A failure only costs the current chunk, and resume (whether from a clean retry or a killed process) naturally continues from the exact byte offset on disk, computed fresh each iteration.
- After a download completes, it verifies the file's size and MD5 checksum against the metadata Zenodo already provides (ZENODO_FILES_DATA) before atomically renaming it into place — a mismatch triggers a retry from scratch rather than silently keeping corrupt data.
- Non-local (remote) destinations keep the old simple atomic-write path, since Range-based resume needs a local temp file to append to.

Bonus fix: while testing this, I found the existing integration test's httpserver mock was dead code  - download_worldcereal_data always used the hardcoded production zenodo.org URL, never the local test server, so the test either hangs (no network) or would silently download real multi-GB files from Zenodo (with network). I patched [tests/integration/data_sources/test_worldcereal.py](vscode-webview://143cvqcfvhq7lpq26lp5jq3tomo95jgtc122h7u5vucsharumlaq/tests/integration/data_sources/test_worldcereal.py) to point the download link at the local httpserver and to compute matching size/checksum for the generated fixture, so the test now actually exercises the download path.


robmarkcole:
Address the Copilot comments:

- Remote-fs fallback (confirmed bug) — it previously wrote straight to the destination with zero retry and zero size/checksum verification. Refactored so both local and remote paths share one _download_to_local_path_with_resume core; for remote destinations, the download is now staged to a local temp file, fully verified there, and only copied to the remote path once confirmed correct — so a bad transfer can never publish a corrupt object that zip_filepath.exists() would then skip forever.
- Double retry budget (confirmed bug) — create_retry_session()'s own 5 retries were nested inside my 5-attempt chunk loop, risking ~25 requests with compounding backoff on a persistent failure. Now the session's automatic retries are disabled (total_retries=0), and the one chunk-retry loop handles both exceptions and retriable HTTP statuses (429/500/502/503/504) through a single MAX_DOWNLOAD_ATTEMPTS counter.
- Exact-size-partial reset (stale) — already fixed as a side effect of the bounded-chunk rewrite in the prior commit; no action needed.
- Test mutating class state (confirmed bug) — _setup_worldcereal_httpserver now monkeypatches a deep copy of WorldCereal.ZENODO_FILES_DATA instead of mutating the shared class list in place, so production metadata is restored after each test. 
- Missing resume/corruption test coverage — added three tests: resuming from a pre-seeded partial file (asserts the Range request targets the exact prior offset), checksum-mismatch cleanup (raises + deletes the corrupt .partial), and the new remote-staging path (successful publish + a mismatch that must not leave a corrupt object at the remote destination).

favyen2:
Thanks for these improvements and fixes! One question before I look more closely, do you think we could use a library like pypdl instead of implementing the resumable download stuff ourselves? It's not that much code so it's fine if not, but if we can leverage some off-the-shelf library then that could save us some lines of code and tests.

robmarkcole:
@favyen2 I'm not familar with pypdl. codex on this topic:

**I’d keep the current implementation rather than use pypdl.**

pypdl is capable and maintained, but its model doesn’t quite match our Zenodo workaround:

- It divides the file into a fixed number of segments and downloads all segments concurrently. There’s no separate maximum request size and concurrency limit. [[Source](https://github.com/mjishnu/pypdl/blob/c04db2cfb4244a2654e2052b3c496208d07d47be/pypdl/consumer.py#L117-L129)](https://github.com/mjishnu/pypdl/blob/c04db2cfb4244a2654e2052b3c496208d07d47be/pypdl/consumer.py#L117-L129)
- Its default five segments would still produce multi-gigabyte requests. Limiting segments to 128 MiB would launch about 29–187 simultaneous requests for these archives.
- We would still need wrapper code for expected-size/MD5 verification, corruption cleanup, remote `UPath` staging, and publishing only after validation.
- It would add direct `aiohttp` and `aiofiles` dependencies. [[Source](https://github.com/mjishnu/pypdl/blob/c04db2cfb4244a2654e2052b3c496208d07d47be/setup.py#L27-L27)](https://github.com/mjishnu/pypdl/blob/c04db2cfb4244a2654e2052b3c496208d07d47be/setup.py#L27-L27)
- We could remove some low-level tests, but would retain most adapter-level tests covering checksum failure and remote destinations.

The current sequential range loop is straightforward and precisely captures the unusual requirement. pypdl would make sense if it exposed `max_segment_size` plus bounded segment concurrency.

rslearn/data_sources/worldcereal.py
```diff
@@ -337,6 +387,169 @@ def filepath_for_product_aez(path_to_tifs: UPath, aez: int) -> UPath | None:
...
+        the underlying session has its own automatic retries disabled, so a
+        persistent failure can't multiply into two nested retry loops.
+
+        Args:
+            file_url: the URL to download from.
+            local_path: the local destination path for the completed download.
+            expected_size: the expected size of the file in bytes.
+            expected_checksum: the expected MD5 checksum of the file.
+        """
+        session = create_retry_session(total_retries=0)
```
favyen2: Should it just use a vanilla Session here if retry within the session is not needed?

favyen2 (approved):

rslearn/data_sources/worldcereal.py
robmarkcole: True, replaced with `session = requests.Session()`

PR #707, opened 2026-09-17
Add components that are needed for the OlmoEarth LCC model.

favyen2:
Also update Concatenate so it retains the timestamps so that downstrean we can make use of it (e.g. in the OlmoEarth model when operating over concatenated inputs).


rslearn/models/temporal_transformer.py
```diff
@@ -0,0 +1,90 @@
...
+
+import torch
+from einops import rearrange
+from torch import nn
+
+from rslearn.models.component import IntermediateComponent, TokenFeatureMaps
+from rslearn.train.model_context import ModelContext
+
+
+class TemporalTransformer(IntermediateComponent):
```
gabrieltseng, line 13: I assume this gets added onto the output of a ViT encoder? I'm curious if this makes a big difference since the whole model is getting finetuned

gabrieltseng (approved):
It would be nice to add a comment / note describing how these all get wired together in the LCC model

favyen2:
Added notes above about how they are used in LCC. Will add some notes to the code to document how they could be used in general.

rslearn/models/temporal_transformer.py
favyen2: added a note that it is most useful for encoder that doesn't reason across timesteps.

For OlmoEarth LCC it is used because it inputs 20 timesteps but that is slow with full attention, so it splits up encoding to two sequences of 10 timesteps, then temporal transformer is used to reason across all 20 timesteps afterwards.

rslearn/train/transforms/concatenate.py
```diff
@@ -34,8 +35,11 @@ def __init__(
                 retain, or empty list to use all bands.
             output_selector: the output selector under which to save the concatenate image.
             concatenate_dim: the dimension against which to concatenate the inputs
+            skip_missing: if True, selectors that are absent from the input/target dicts
+                are silently skipped instead of raising. Useful when concatenating
+                optional inputs (e.g. per-month layers where some months are missing).
```
yawenzzzz, line 40: When we do token pooling, OlmoEarth has the missing token masks that used to masked out those tokens before pooling. I am wondering if we can still have the missing mask here, instead of simply skipping missing tokens, which will make samples temporally misaligned.

yawenzzzz: https://github.com/allenai/rslearn/blob/778bcdcc7b4933642f3d0e6c8cc377453d656cfd/rslearn/models/olmoearth_pretrain/model.py#L789

rslearn/models/temporal_transformer.py
```diff
@@ -0,0 +1,93 @@
...
+    ) -> None:
+        """Create a new TemporalTransformer.
+
+        Args:
+            dim: the token embedding dimension C.
+            depth: number of transformer encoder layers.
+            num_heads: number of attention heads.
+            dim_feedforward: hidden size of the feed-forward network in each layer.
+            dropout: dropout rate in the transformer layers.
+            positional_embedding_num_tokens: if set, add a learned positional
```
yawenzzzz, line 44: Same thing here, if the inputs samples are basically temporally misaligned, then will this learned positional embedding still be good for them? also, i guess we can still use the timestamps (like month) based positional encoding here if we carry the timestamps information, and potentially use the `key_padding_mask` to remove the missing timestamps from attention right?

yawenzzzz (approved):

rslearn/models/temporal_transformer.py
favyen2: OK I added mask to TokenFeatureMaps.

favyen2:

